PR #21, opened 2015-01-04
changed 3 tests to also work with structs

masters3d:
Previous tests forced the use of only classes. Now reference and values
types can be tested


grade-school/GradeSchoolTest.swift
```diff
@@ -11,16 +11,16 @@ class GradeSchoolTest: XCTestCase
     }
     
     func testAddStudent() {
-        let school = GradeSchool()
+        var school = GradeSchool() // var
```
kytrinyx: What does the comment mean?

PR #28, opened 2015-02-26
luhn + series + pig-latin + pythagorean-triplet + difference-of-squares

masters3d:


kytrinyx:
Thank you!

kytrinyx:
One thing that caught my eye is that you have several successive blank lines in a few places. Is that intentional? As I've gotten my feet wet in Go I've realised that what seems odd in one language is perfectly natural in another, so I just curious.

masters3d:
That is a good point. I sometimes use the white space to keep me oriented. Maybe not best practice.  Maybe in the future we will get a tool to "lint" all our swift code to a standard. 
http://stackoverflow.com/questions/24215604/is-there-a-canonical-style-for-swift-code

kytrinyx:
Yeah, a linter would be nice.

PR #30, opened 2015-03-04
Invalid response to forceful question in test case

patrikfors:
Imo the response to a forceful question should be "Sure.". The spec isn't perfectly clear but the question requirement comes before the shouting requirement and should have precedence.


masters3d:
Thanks for the pointing this out @patrikfors  In the spec a "forceful question"  response is "Whoa, chill out!"   https://github.com/exercism/x-common/blob/master/bob.json#L53-L55

All other tracks have the same implementation so I am going to close this issue for this swift track. 

If you feel that spec is wrong,  you are welcome to request a change there but I think the current response makes more sense than "sure".

PR #52, opened 2015-06-19
Use enum and struct to model nucleotide and DNA

aarti:
I tried to add some swift approaches to this example. 

1. The DNA should be modeled as a struct since the following apply to it.
*The structure’s primary purpose is to encapsulate a few relatively simple data values.
It is reasonable to expect that the encapsulated values will be copied rather than referenced when you assign or pass around an instance of that structure.
Any properties stored by the structure are themselves value types, which would also be expected to be copied rather than referenced.
The structure does not need to inherit properties or behavior from another existing type.*
 
2. The nucleobases(A,C,T,G) are enums, which is sort of self-explanatory. Nice that swift allows enums to be `Character` type.

3. The Optional Bindings `if let` made sense when converting from RawValue of `Character` to `enum`.

4. The DNA `struct` has a failable initializer created by adding `?` to `init` and returns `nil` if the strand has an invalid character.

### DIFF
--- a/nucleotide-count/NucleotideCountTest.swift
+++ b/nucleotide-count/NucleotideCountTest.swift
@@ -9,11 +9,10 @@ class NucleotideCountTests: XCTestCase
         XCTAssertEqual(result, expected)
     }
     
-    // An initializer defined with init can be made failable by adding a ? or a ! after the init
     func testEmptyDNAStringHasNoNucleotides() {
         let dna = DNA(strand: "")!
-        let results = dna.nucleotideCounts
-        let expected:[Nucleobase:Int] = [ .Adenine: 0, .Thymine: 0, .Cytosine: 0 , .Guanine : 0 ]
+        let results = dna.counts()
+        let expected = ["T": 0, "A": 0, "C": 0, "G": 0]
         XCTAssertEqual(results, expected)
     }
     
@@ -26,8 +25,8 @@ class NucleotideCountTests: XCTestCase
     
     func testRepetitiveSequenceHasOnlyGuanosine() {
         let dna = DNA(strand: "GGGGGGGG")!
-        let results = dna.nucleotideCounts
-        let expected:[Nucleobase:Int] = [ .Adenine: 0, .Thymine: 0, .Cytosine: 0 , .Guanine : 8 ]
+        let results = dna.counts()
+        let expected = [ "A": 0, "T": 0, "C": 0 , "G": 8 ]
         XCTAssertEqual(results, expected)
     }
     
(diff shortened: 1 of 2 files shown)


nucleotide-count/NucleotideCountTest.swift
```diff
@@ -1,60 +1,60 @@
...
         XCTAssertEqual(result, expected)
     }
     
     func testCountsANucleotideOnlyOnce() {
-        let dna = DNA.withStrand("CGATTGGG")!
-        var result = dna.count("T")!
-        result = dna.count("T")!
+        let dna = DNA(strand: "CGATTGGG")!
+        var result = dna.count("T")
+        result = dna.count("T")
```
hankturowski: What is the purpose of this line?

hankturowski:
1. Remember that the example file just needs to prove that the tests will pass. You're right that it should be a struct, but end users never see the example files. Having said that, it's fantastic practice to make updates like this and there's no problem in doing so.

2. The only issue I see here is that by insisting on an an enum for the nucleobases, you're letting your implementation details leak. It's normally bad design to insist that the consumer of your API use your internal representation. For all we know, the consumer is reading text from a text file and needs plain text back. If we're going to do this then at the very least we should provide the enum in the test file and tell the user to use it, but I think we should stick to characters to stay in line with the rest of the languages.

3. Optional binding is a good thing.

4. Using failable initializers is good here. The original version was written before they existed in the language. Nice catch.

nucleotide-count/NucleotideCountTest.swift
aarti: It's redundant

aarti:
Pretty standard to use enum's in Unit Tests. Java has had them for a while, back when I wrote in Java. Otherwise we can write a map function to serialize to a generic format to match the other tests

masters3d:
2. I like the short string vs longer enum

aarti:
Enum vs String is a frequently debated topic. I think the goal of exercism is not simply that we keep api of the tests consistent but that we effectively program in the language we are in( in the most idiomatic way). To that end I think this test should use enum's.  @kytrinyx 
http://stackoverflow.com/questions/277364/java-enum-vs-string-as-parameters

hankturowski:
I think I wasn't clear before.

Using enum is an elegant way to SOLVE the exercise, but we don't want to push the user into one specific solution. We want the user to be able to uncover the solution for themselves, and get feedback on it from others. Showing them the enum from the beginning is only giving them one real way to complete the exercise.

aarti:
I updated the test to keep the return type generic. Internally struct uses an enum, but I added a new method that returns a dictionary of `[String: Int]`
- counts() which returns this generic format
It is also goes along with the original count method
- count("A") returns value for one
- counts() returns all

hankturowski:
This looks great @aarti. I really do like your enum solution. It's very clean. I encourage you to keep an eye on this exercise as people complete it on the site, and give them feedback as this type of feedback is really helpful to users.

aarti:
Thanks. I will do that, its a difficult language to learn, so many ways to model things.

PR #57, opened 2015-08-13
Moving bob down the list of exercises

hankturowski:
Bob is currently a roadblock for many users. Moving it down the list.


masters3d:
Thanks

kytrinyx:
Good call, thanks!

PR #68, opened 2015-09-16
update binary to return Optionals

masters3d:
https://github.com/exercism/x-common/issues/95

### DIFF
--- a/binary/BinaryExample.swift
+++ b/binary/BinaryExample.swift
@@ -7,7 +7,7 @@ struct Binary {
     
     var toDecimal:Int {get {return Int(UIntValue)}}
     
-    private func bi2Uint(input:String) -> UInt{
+    private func bi2Uint(input:String) -> UInt?{
         let orderedInput = Array(input.characters.reverse())
         var tempUInt:UInt = 0
         for (inx,each) in orderedInput.enumerate(){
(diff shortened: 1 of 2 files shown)


hankturowski:
Your Travis integration is an amazing improvement.

hankturowski:
I like this update a lot. We should do more with optionals.

masters3d:
Thanks! Travis is cool. takes about 2-3 minutes to run though, not sure if that is normal. 
I was debating between throwing or optionals but I am not a fan of making all the callers have to use try try? for everything. 
http://ericasadun.com/2015/07/29/nils-vs-throws-in-swiftlang-part-ii/

hankturowski:
I think that 2-3 minutes is a small price to pay for the added layer of protection. With the test running sequentially, I'm not surprised it takes that long.

I agree with you on try?. We should look through the exercises and see which ones might be good candidates for try blocks. Maybe we set up some specific error types to be returned? I kind of like the idea of an exercise that defines some things ahead of time and expects the user to conform. Like they're extending an existing system and can't touch the existing error types.

Maybe we can push a starter exercise.swift file with a predefined class and set of errors, then ask the user to extend it without touching the existing code.

kytrinyx:
Travis might speed up if we configure it with `sudo: false` so that it uses containerized builds.

masters3d:
i did try ```sudo: false``` on a test branch but not much improvement. I can live with 2-3 mins wait per pr. ;)

PR #72, opened 2015-09-29
Fix typo

robtimp:
Scrambble -> Scrabble


kytrinyx:
Nice catch, thanks :)

PR #73, opened 2015-10-03
Fix typo

robtimp:
Zelot --> Zealot


kytrinyx:
Good catch, thanks :)

PR #76, opened 2015-11-21
Added Custom Set Problem

masters3d:


kytrinyx:
Would you mind rebasing onto the up-to-date master to get rid of that merge commit?

Other than that this looks good to me. That said, I'm no Swift developer. @genevievelesperance @marianaIAm want to give this a second look?

hankturowski:
Unless there's some Array weirdness in there that I'm not seeing, we should use `[Int]()` instead of `Array<Int>()` to conform to Apple's preferred style. I'm away from my computer or I'd make the change myself.

masters3d:
I'll update and rebase this on another branch. I don't git git. ;)

PR #81, opened 2016-01-09
Add Nth Prime exercise

robtimp:
Add Nth Prime exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -14,7 +14,7 @@
     "phone-number",
     "grade-school",
     "robot-name",
-     "perfect-numbers",
+    "perfect-numbers",
     "allergies",
     "accumulate",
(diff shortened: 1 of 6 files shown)


masters3d:
Cool, could you add this exercise to the Xcode project so that Travis can can run?
Maybe later we can update Travis to use Xcode workspaces so files get added automatically to Travis.

robtimp:
@masters3d No problem: I completely missed that there was an Xcode project at all!

masters3d:
Nice. Yeah. I have been meaning to add it to the document section of the readme. Maybe once we make it Linux compatible when swift 3 comes out, we could come up with a different way for Travis to run the unit tests.

masters3d:
Thank you.

PR #83, opened 2016-01-10
Update GradeSchool{Example,Test}

yurrriq:
Test:

- Fix minor formatting issues
- Add generic `sameStudents(_:_:)` to allow more flexibility in solutions
- Avoid unnecessarily forced coercion
- `s/db/roster/g; s/sortedRoster\(\)/sortedRoster/g;`

Example:

- Use a `struct` instead
- Rename `db` to `roster`
- Make `sortedRoster` a computed property, rather than a function
- Use nil coalescence for succinct `studentsInGrade(_:`) definition
- Use appropriately sized empty dictionary in `sortedRoster`,
  instead of self-mutation
- Don't replace `[String]` with `Set<String>`, though now users can :smile:

### DIFF
--- a/grade-school/GradeSchoolTest.swift
+++ b/grade-school/GradeSchoolTest.swift
@@ -2,82 +2,89 @@ import XCTest
 
 // Apple Swift version 2.0
 
-class GradeSchoolTest: XCTestCase
-{
-
+class GradeSchoolTest: XCTestCase {
     func testAnEmptySchool() {
-        let school = GradeSchool()
+        let school   = GradeSchool()
         let expected = [:]
-        let result = school.db
+        let result   = school.roster
         XCTAssertEqual(result, expected)
     }
-    
+
     func testAddStudent() {
-        let school = GradeSchool()
+        var school = GradeSchool()
         school.addStudent("Aimee", grade: 2)
-        let result = school.db
+        let result = school.roster
         let expected: Dictionary = [2: ["Aimee"]]
         XCTAssertEqual(Array(result.keys), Array(expected.keys))
-        XCTAssertEqual(result[2]!, expected[2]!)
+        XCTAssert(sameStudents(result[2], expected[2]))
     }
 
     func testAddMoreStudentsInSameClass() {
-        let school = GradeSchool()
+        var school = GradeSchool()
         school.addStudent("Fred", grade: 2)
         school.addStudent("James", grade: 2)
         school.addStudent("Paul", grade: 2)
-        let result = school.db
+        let result   = school.roster
         let expected = [2: ["Fred", "James", "Paul"]]
         XCTAssertEqual(Array(result.keys), Array(expected.keys))
-        XCTAssertEqual(result[2]!, expected[2]!)
+        XCTAssert(sameStudents(result[2], expected[2]))
     }
 
     func testAddStudentsToDifferentGrades() {
-        let school = GradeSchool()
-        school.addStudent("Chelsea", grade: 3)
+        var school = GradeSchool()
+        school.addStudent("Chelsea", grade:3)
         school.addStudent("Logan", grade: 7)
-        let result = school.db
+        let result = school.roster
         let expected = [3: ["Chelsea"], 7: ["Logan"]]
         XCTAssertEqual(Array(result.keys).sort(>), Array(expected.keys).sort(>))
-        XCTAssertEqual(result[3]!, expected[3]!)
+        XCTAssert(sameStudents(result[3], expected[3]))
     }
 
     func testGetStudentsInAGrade() {
-        let school = GradeSchool()
+        var school = GradeSchool()
         school.addStudent("Franklin", grade: 5)
         school.addStudent("Bradley", grade: 5)
         school.addStudent("Jeff", grade: 1)
-        let result = school.studentsInGrade(5)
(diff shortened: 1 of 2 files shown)


grade-school/GradeSchoolTest.swift
```diff
@@ -2,82 +2,89 @@ import XCTest
...
+
     func testAddStudent() {
-        let school = GradeSchool()
+        var school = GradeSchool()
         school.addStudent("Aimee", grade: 2)
-        let result = school.db
+        let result = school.roster
         let expected: Dictionary = [2: ["Aimee"]]
         XCTAssertEqual(Array(result.keys), Array(expected.keys))
-        XCTAssertEqual(result[2]!, expected[2]!)
```
masters3d: I think assert equal is the way to go here but I do agree that we should not force unwrap these optionals.  There was an issue early in xctesf where xct assert equals did not work on optionals but I believe now it does.

grade-school/GradeSchoolTest.swift
```diff
@@ -2,82 +2,89 @@ import XCTest
...
         XCTAssertEqual(Array(result.keys).sort(>), Array(expected.keys).sort(>))
-        XCTAssertEqual(result[3]!, expected[3]!)
-        XCTAssertEqual(result[4]!, expected[4]!)
-        XCTAssertEqual(result[6]!, expected[6]!)
+        XCTAssert(sameStudents(result[3], expected[3]))
+        XCTAssert(sameStudents(result[4], expected[4]))
+        XCTAssert(sameStudents(result[6], expected[6]))
+    }
+
+    func sameStudents<C: CollectionType where C.Generator.Element == String>(result: C?, _ expected: [String]?) -> Bool {
```
masters3d: I don't like introducing this function here.  xCTAsssertEquals should be able to test optional. I rather remove the unwrapping than introducing this function. You may need to implement the equality protocol in the example file.

yurrriq:
I'm out for the night and was gonna work up another solution later, but have you changed your mind?

The main reason behind the added function was to allow for implementations that use sets instead of arrays. I'm not married to this solution, but it gets the job done.

Related: Do you have a preferred method for keeping giant function signatures within e.g. 100 characters? I haven't found my favorite yet, but I tend to lean toward a Haskelly approach with parens and commas starting lines.

masters3d:
I was more concern about being able to compare optional arrays/sets http://stackoverflow.com/questions/29478665/comparing-optional-arrays

It may be a little cleaner if you just extend xctest with a func like XCTAssertEqualOptionalArray 

maybe typealias could help shorten some signatures but I don't recommend it. 
https://medium.com/swift-programming/swift-typealias-to-the-rescue-b1027fc571e3#.dyxt451xw

yurrriq:
We could do that and take out the `guard let`. What's the harm in comparing optional a with `XCAssertEqual`? False positives when `nil == nil`? That shouldn't be an issue here anyway, so maybe we could pull it off without the extension..

yurrriq:
Totally unrelated, but thanks for passively introducing me to `var let`. I was previously unaware and I'm sure it'll come in handy sometime.

yurrriq:
Re: `typealias` to the rescue. Can you make aliases for constrained generic types? I don't think so, but I'd love to learn otherwise.

masters3d:
I don't think you can create generic typealias

I was thinking something like this

```swift 
private extension XCTestCase {
   func XCTAssertEqual <T : Equatable>(collection1 : [T]? , collection2 : [T]? ) {
       XCTAssert(sameStudents(collection1, collection2))
}}

masters3d:
you may not even need the sameStrudent function if you use something along these lines.

```swift
func ==<T: Equatable>(lhs: [T]?, rhs: [T]?) -> Bool {
    switch (lhs,rhs) {
    case (.Some(let lhs), .Some(let rhs)):
        return lhs == rhs
    case (.None, .None):
        return true
    default:
        return false
    }
}

private extension XCTestCase {
   func XCTAssertEqual <T : Equatable>(collection1 : [T]? , collection2 : [T]? ) {
       XCTAssert(collection1 == collection2))
}}

yurrriq:
#85 forces users to implement their solution using `[Int: [String]]`, instead of, for example, `[Int: Set<String>]`. The change breaks [my solution](http://exercism.io/submissions/56ba320e7dcb4ccb87f53acb5869c928) and basically undermines the whole point of #83 in the first place.

If your intention is to force `[Int: [String]]` then I guess we can agree to disagree. Otherwise, #85 needs to be amended.

masters3d:
The example struct you summited used this signature and not a set. I agree
that we should allow sets to be used. You can't cast a generic swift
dictionary to a NSDictionary. AFAIK, you can't create generic empty
dictionaries pure swift arrays. The limitation is in the first test, xctest
wants a reference file. Maybe we can just do without the first test on
empty dictionaries.

yurrriq:
Right, I didn't want to change too much all at once. :smile: Plus, it was a nice way to validate that the tests work for both `Array`s and `Set`s. Hmm, I'll have to think on this some or see what you come up with.

yurrriq:
Could `.isEmpty` help?

masters3d:
That will work.

I can fix this tomorrow if you don't get to it first.

Thanks for the suggestion. Most of the excersises were implement before
Set<> existed in the language.

PR #92, opened 2016-01-18
Robot Simulator

robtimp:
Add Robot Simulator exercise

### DIFF
--- a/robot-simulator/RobotSimulatorExample.swift
+++ b/robot-simulator/RobotSimulatorExample.swift
@@ -12,6 +12,7 @@ enum CardinalDirection {
 }
 
 class SimulatedRobot {
+    
     var bearing: CardinalDirection = .North
     var x: Int = 0
     var y: Int = 0
@@ -55,15 +56,16 @@ class SimulatedRobot {
             case .West:  x -= 1
         }
     }
-}
-
-enum RobotInstruction: String {
-    case TurnLeft  = "L"
-    case TurnRight = "R"
-    case Advance   = "A"
+    
 }
 
 struct Simulator {
+    
+    enum RobotInstruction: String {
+        case TurnLeft  = "L"
+        case TurnRight = "R"
+        case Advance   = "A"
+    }
     
     func instructions(instructions: String) -> [RobotInstruction] {
         var result = [RobotInstruction]()
--- a/robot-simulator/RobotSimulatorExample.swift
+++ b/robot-simulator/RobotSimulatorExample.swift
@@ -2,51 +2,60 @@
 
 // Apple Swift version 2.1
 
-enum CardinalDirection {
-    case North
-    case East
-    case South
-    case West
+struct SimulatedRobot {
     
-    static let allValues: [CardinalDirection] = [.North, .East, .South, .West]
-}
-
-class SimulatedRobot {
+    enum Direction {
+        case North
+        case East
+        case South
+        case West
+        
+        static let allValues: [Direction] = [.North, .East, .South, .West]
+    }
     
-    var bearing: CardinalDirection = .North
+    enum RobotInstruction: String {
+        case TurnLeft  = "L"
+        case TurnRight = "R"
+        case Advance   = "A"
+    }
+    
+    var bearing: Direction = .North
     var x: Int = 0
     var y: Int = 0
     var coordinates: [Int] {
         return [x, y]
     }
     
-    func orient(bearing: CardinalDirection) {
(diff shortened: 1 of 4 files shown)


robot-simulator/RobotSimulatorTest.swift
```diff
@@ -0,0 +1,150 @@
...
+        simulator.evaluate(robot, commands: "RLAALAL")
+        XCTAssertEqual([0, 2], robot.coordinates)
+        XCTAssertEqual(CardinalDirection.West, robot.bearing)
+    }
+    
+    func testInstructManyRobots() {
+        let robot1 = SimulatedRobot()
+        let robot2 = SimulatedRobot()
+        let robot3 = SimulatedRobot()
+        simulator.place(robot1, x: 0, y: 0, direction: .North)
```
masters3d: If the solution was written with a struct, would these need to be ```var```

robot-simulator/RobotSimulatorExample.swift
```diff
@@ -0,0 +1,97 @@
...
+
+enum RobotInstruction: String {
+    case TurnLeft  = "L"
+    case TurnRight = "R"
+    case Advance   = "A"
+}
+
+struct Simulator {
+    
+    func instructions(instructions: String) -> [RobotInstruction] {
```
masters3d: This is private right?

robot-simulator/RobotSimulatorExample.swift
```diff
@@ -0,0 +1,97 @@
...
+            // See https://github.com/apple/swift-evolution/blob/master/proposals/0004-remove-pre-post-inc-decrement.md
+            case .North: y += 1
+            case .East:  x += 1
+            case .South: y -= 1
+            case .West:  x -= 1
+        }
+    }
+}
+
+enum RobotInstruction: String {
```
masters3d: You can throw this inside the simulator if you are not using it outside.

robot-simulator/RobotSimulatorExample.swift
robtimp: It can't be made private since it's tested directly.

robot-simulator/RobotSimulatorExample.swift
```diff
@@ -0,0 +1,97 @@
...
+        for character in characters {
+            if let instruction = RobotInstruction(rawValue: character) {
+                result.append(instruction)
+            }
+        }
+        
+        return result
+    }
+    
+    func place(robot: SimulatedRobot, x: Int, y: Int, direction: CardinalDirection) {
```
masters3d: Place and evaluate reference to a robot object. This makes think that theses methods should live inside the robot type instead. You can make it a protocol if you want but I don't think you need two instance-able objects. Robots should be sufficient.

robot-simulator/RobotSimulatorExample.swift
masters3d: Instructions is tested? I didn't see it.

robot-simulator/RobotSimulatorTest.swift
robtimp: But if we support `struct`s, wouldn't we also need to use `inout` params, so we're not just updating a copy of the robot?

robot-simulator/RobotSimulatorExample.swift
robtimp: Ok, will do!

robot-simulator/RobotSimulatorExample.swift
robtimp: (see `testInstructionForTurningLeft()`, etc.)

robot-simulator/RobotSimulatorTest.swift
masters3d: Yea if you keep the simulator struct methods.  If the tests work with structs then they will also work with classes.

robot-simulator/RobotSimulatorExample.swift
masters3d: Ah, I see. This forces people to use another enum. I think the best approach here would be to go only use two types. One enum and another struct/class for the solution.

masters3d:
Thanks for the PR. Before I can accept could you change it a bit?

- move the solution to one struct/class
- combine the enums a into one ( you can have methods in a enum a too )

robtimp:
@masters3d Ok, made the updates. The tests will now work with either a struct or class.

masters3d:
Looks good!

PR #95, opened 2016-01-20
Add Queen Attack

robtimp:
Add Queen Attack exercise.

### DIFF
--- a/queen-attack/QueenAttackExample.swift
+++ b/queen-attack/QueenAttackExample.swift
@@ -1,0 +1,66 @@
+import Foundation
+
+// Apple Swift version 2.1
+
+struct Queens {
+    
+    enum InitError: ErrorType {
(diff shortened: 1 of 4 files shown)


queen-attack/QueenAttackExample.swift
```diff
@@ -0,0 +1,66 @@
+// Foundation not needed
+
+// Apple Swift version 2.1
+
+struct Queens {
+    
+    enum InitError: ErrorType {
+        case SameSpace
+        case IncorrectNumberOfCoordinates
+        case InvalidCoordinates
```
masters3d: Are you going to test the other two cases?

robtimp: Ok, good idea!

masters3d:
Nice! Thanks for all the good work.

PR #97, opened 2016-01-21
Moving exercises to a subfolder

masters3d:


masters3d:
https://github.com/exercism/xswift/issues/96

PR #100, opened 2016-01-21
Add Binary Search

robtimp:


masters3d:
Looks good but could you redo the pull request with just one commit? I don't know git that well but I've notice when I've merge duplicating commits that those files then have merge conflicts. 

Sometimes you can go to the last commit on the remote origin and check it out locally. I'm going to close this one. (I know there is a way to squash commits on the remote but I haven done that before.)

kytrinyx:
This is the process @robtimp could use to fix this pull request instead of opening a new one.
http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

If you want to do this yourself, @masters3d, then you can say:

```
git fetch origin refs/pull/100/head
git checkout -b binary-search FETCH_HEAD
```

Then squash locally, which should make @robtimp the author and you the committer.

Then you can switch to master and `git merge --no-ff binary-search`, and change the merge message to something like _Merge pull request #100 from robtimp/binary-search_.

masters3d:
That was very cool. Thanks for the instructions! @kytrinyx 
It wasn't as scary as I thought it was going to be an I even learned some vi commands. =)

kytrinyx:
You can also set the editor as an environment variable--can't remember what it is though. GIT_EDITOR? Just EDITOR? Not sure.

masters3d:
ah interesting. 
http://stackoverflow.com/questions/2596805/how-do-i-make-git-use-the-editor-of-my-choice-for-commits
Once I knew how to exit recoding and insert mode, I was back on business. Not bad. I need to learn Vim at some point.

kytrinyx:
I feel like on a scale from 1 to 10 "I should learn vim sometime" falls around a 2. I mean I actually do use vim, but that's immaterial. If you have an editor that works for you, then changing editors is going to have to give you a pretty serious bump in productivity for it to be worth the hit in productivity to actually learn to use it.

PR #102, opened 2016-01-27
Add Binary Search Tree exercise

robtimp:
@masters3d I don't believe a solution is currently possible with structs (I get an error that `Recursive value type BinarySearchTree<T> is not allowed` when I try), but I'm using `var` instead of `let` just in case. 

Any thoughts on that?

### DIFF
--- a/config.json
+++ b/config.json
@@ -55,7 +55,8 @@
     "kindergarten-garden",
     "robot-simulator",
     "queen-attack",
-    "binary-search"
+    "binary-search",
+    "binary-search-tree"
   ],
   "deprecated": [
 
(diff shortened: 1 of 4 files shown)


masters3d:
I'll take a look this evening. I know that for the linked list exercises I had to use a class eventhough a struct based solution was possible but it required jumping through many hoops.

robtimp:
@masters3d Could also probably be done with an enum, using `indirect`, but a class is definitely the simplest solution.

masters3d:
waynewbishop.com/swift/binary-search-trees/

Looks good, it's just a glorified Linked list. Thanks.

masters3d:
Ah yeah, the new indirect enums would do it.

PR #103, opened 2016-01-28
Add Hexadecimal exercise

robtimp:


masters3d:
You are on a roll!
Instead of using the method .toDecimal extend the Int value to take a hex object. 

https://github.com/exercism/xswift/commit/ee41be07b31a85fadb43c568b41acb8259c7af2f

I thought about making the custom number object equatable to Int but that would be too magical.

robtimp:
@masters3d Thanks! 

I checked out your example and updated to Initialize an Int from the Hexadecimal object.

masters3d:
Thanks!

PR #106, opened 2016-01-30
Revert "update isosceles test"

masters3d:
Reverts exercism/xswift#105


masters3d:
https://github.com/exercism/xswift/issues/104

kytrinyx:
I'm tempted to suggest taking out the edge case, since the distinction between degenerate triangles and "not a triangle" is so subtle, and kind of beside the point here. What do you think?

masters3d:
I agree with this.

PR #107, opened 2016-01-31
Remove edge case 2-4-2 triangle

masters3d:


masters3d:
https://github.com/exercism/xswift/issues/104

PR #108, opened 2016-02-01
Add Largest Series Product exercise

robtimp:


exercises/largest-series-product/LargestSeriesProductExample.swift
```diff
@@ -0,0 +1,51 @@
+// Foundation not needed
```
masters3d: Do you need foundation here? I am just trying to prepare for Open Source Swift so we can run and linux and most of the Open Foundation classes have not been implemented.

robtimp: No, I just took it out.

masters3d:
Thanks!

PR #112, opened 2016-02-04
Add Minesweeper exercise

robtimp:
@masters3d Are you ok including `Foundation` for this one? Otherwise the border validation will be trickier.


masters3d:
Yes for regex. Hopefully we get good support later with open foundation. We can always change later.

PR #113, opened 2016-02-05
Add OCR Numbers exercise

robtimp:
@masters3d I've attempted to improve on the Ruby version by validating the input has a correct number of lines and columns.

### DIFF
--- a/exercises/ocr-numbers/OcrNumbersTest.swift
+++ b/exercises/ocr-numbers/OcrNumbersTest.swift
@@ -1,0 +1,189 @@
+import XCTest
+
+// Apple Swift version 2.1
+
+class OcrNumbersTest: XCTestCase {
+    
+    func testRecognizeZero() {
+        let text =  " _ \n" +
+                    "| |\n" +
+                    "|_|\n" +
+                    "   "
+        XCTAssertEqual("0", try? OCR(text).converted())
+    }
+    
+    func testRecognizeOne() {
+        let text =  "   \n" +
+                    "  |\n" +
+                    "  |\n" +
+                    "   "
+        XCTAssertEqual("1", try? OCR(text).converted())
+    }
+
+    func testRecognizeTwo() {
+        let text =  " _ \n" +
+                    " _|\n" +
+                    "|_ \n" +
+                    "   "
+        XCTAssertEqual("2", try? OCR(text).converted())
+    }
+    
+    func testRecognizeThree() {
+        let text =  " _ \n" +
+                    " _|\n" +
+                    " _|\n" +
+                    "   "
+        XCTAssertEqual("3", try? OCR(text).converted())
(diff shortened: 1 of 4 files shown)


exercises/ocr-numbers/OcrNumbersTest.swift
```diff
@@ -0,0 +1,187 @@
...
+        
+        do {
+            let text =
+                "    _\n" +
+                "  | _\n" +
+                "  ||_\n" +
+                "     "
+            let _ = try OCR(text)
+        } catch OCR.Error.InvalidNumberOfColumns {
+            throwsInvalidNumberOfColumnsError = true
```
masters3d: Can you return here after this assignment?

exercises/ocr-numbers/OcrNumbersTest.swift
```diff
@@ -0,0 +1,187 @@
...
+// Apple Swift version 2.1
+
+class OcrNumbersTest: XCTestCase {
+    
+    func testRecognizeZero() {
+        let text =  " _ \n" +
+                    "| |\n" +
+                    "|_|\n" +
+                    "   "
+        XCTAssertEqual("0", try? OCR(text).convert())
```
masters3d: Let's call this 'converted' if is not mutating to start following the new naming guidelines. (Well need to revise the rest of the problems once that comes out. )

robtimp: Ah, good call!

robtimp:
@masters3d Ok, updated

exercises/ocr-numbers/OcrNumbersTest.swift
```diff
@@ -0,0 +1,189 @@
...
+        
+        do {
+            let text =
+                "    _  _ \n" +
+                "  | _| _|"
+            let _ = try OCR(text)
+        } catch OCR.Error.InvalidNumberOfLines {
+            throwsInvalidNumberOfLinesError = true
+            return
+        } catch {
```
masters3d: Do you need two catches here? :)

exercises/ocr-numbers/OcrNumbersTest.swift
masters3d: Maybe I am missing something about catching.

masters3d: I get it, you are catching any other error. Mmm. Let's go for it.

masters3d:
Thanks. Looks good. I wish swift error catching was a little more elegant.

exercises/ocr-numbers/OcrNumbersTest.swift
robtimp: Yep, `catch` statements have to be exhaustive.

PR #114, opened 2016-02-07
Add Palindrome Products exercise

robtimp:
@masters3d A bit of Swift oddness I noticed while working on this one: I apparently can't use `.contains()` with `[[Int]]` (an array of arrays of ints), but I _can_ loop over the array and use `==` to see if a given `[Int]` is in there. Strange!

I guess this has something to do with `Array<Int>` being Equatable, but not Hashable?

### DIFF
--- a/exercises/palindrome-products/PalindromeProductsExample.swift
+++ b/exercises/palindrome-products/PalindromeProductsExample.swift
@@ -1,0 +1,72 @@
+// Foundation not needed
+
+// Apple Swift version 2.1
+
+struct PalindromeCollection: CollectionType {
+    
+    typealias Index = Int
+    
+    var startIndex: Int {
+        return 0
+    }
+    
+    var endIndex: Int {
+        return count
+    }
+    
+    subscript(i: Int) -> Palindrome {
+        return contents[i]
+    }
+    
+    typealias Palindrome = (value:Int, factors: [[Int]])
+    
+    private let contents: [Palindrome]
+    
+    let largest: Palindrome
+    let smallest: Palindrome
+    
+    init(maxFactor: Int, minFactor: Int = 1) {
+        let result = PalindromeCollection.generate(maxFactor: maxFactor, minFactor: minFactor)
+        
+        self.contents = result
+        self.largest = result.last!
+        self.smallest = result.first!
+    }
+    
+    private static func generate(maxFactor maxFactor: Int, minFactor: Int) -> [Palindrome] {
+        var palindromes = [Int : Palindrome]()
+        
+        for i in minFactor...maxFactor {
+            for j in minFactor...maxFactor {
+                let product = i * j
+                
+                if numberIsPalindrome(product) {
+                    let newFactor = [i, j].sort()
+                    var factors = [newFactor]
+                    
+                    if let currentPalindrome = palindromes[product] {
+                        for factor in currentPalindrome.factors {
+                            if factor == newFactor {
+                                continue
+                            }
+                            factors.append(factor)
+                        }
+                    }
+                    
+                    palindromes[product] = Palindrome(value: product, factors: factors)
+                }
+            }
+        }
+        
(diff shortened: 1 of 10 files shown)


exercises/palindrome-products/PalindromeProductsExample.swift
```diff
@@ -0,0 +1,60 @@
+// Foundation not needed
+
+// Apple Swift version 2.1
+
+struct Palindromes {
+    
+    struct Palindrome {
```
masters3d: I don't see why this struct needs to be nested. My preference would be to only have 'Palindrome' as the main struct with largest and smallest as Int properties.

masters3d:
What do you think of renaming `Palindromes` to `PalindromeSequence` and having it conform to the sequence protocol? 
https://www.natashatherobot.com/swift-conform-to-sequence-protocol/

exercises/palindrome-products/PalindromeProductsExample.swift
masters3d: I take that back. You do need two different structs but we should conform one to the sequence type.

masters3d:
The issue is an Array<Array<Int>>is too generic since array itself is generic. I bet is just a compiler limitation.

robtimp:
@masters3d I've been looking into that post and the sequence protocol, but I haven't been able to figure out a way to make it work here. 

You can use the Sequence protocol to do a for/in loop, but how would you get the largest and smallest palindromes? Could you give me some more guidance what you have in mind?

masters3d:
you are right, I guess I was thinking Collection type instead of Sequence

this needs some refactoring but this is what I was thinking. 

```Swift

// Foundation not needed

// Apple Swift version 2.1

struct PalindromeCollection:CollectionType {
    
    typealias Index = Int
    
    var startIndex: Int {
        return 0
    }
    
    var endIndex: Int {
        return count
    }
    
    subscript(i: Int) -> (value:Int, factors: [[Int]]) {
        return (contents[i].value, contents[i].factors)
    }
    
    typealias Palindrome = (value:Int, factors: [[Int]])

    private var contents = [Palindrome]()
    
    let largest: Palindrome
    let smallest: Palindrome
    
    init(maxFactor: Int, minFactor: Int = 1) {
        let result = PalindromeCollection.generate(maxFactor: maxFactor, minFactor: minFactor)
        
        self.contents = result
        self.largest = result.last!
        self.smallest = result.first!
    }
    
    private static func generate(maxFactor maxFactor: Int, minFactor: Int) -> [Palindrome] {
        var palindromes = [Int : Palindrome]()
        
        for i in minFactor...maxFactor {
            for j in minFactor...maxFactor {
                let product = i * j
                
                if numberIsPalindrome(product) {
                    let newFactor = [i, j].sort()
                    var factors = [newFactor]
                    
                    if let currentPalindrome = palindromes[product] {
                        for factor in currentPalindrome.factors {
                            if factor == newFactor {
                                continue
                            }
                            factors.append(factor)
                        }
                    }
                    
                    palindromes[product] = Palindrome(value: product, factors: factors)
                }
            }
        }
        
        let sorted = palindromes.sort { palindrome1, palindrome2 -> Bool in
            return palindrome1.0 < palindrome2.0
        }

        let toReturn = sorted.map{(value:$0.1.value,factors: $0.1.factors )}
        
        return toReturn
    }
    
    private static func numberIsPalindrome(number: Int) -> Bool {
        let numberString = String(number)
        
        return numberString == String(numberString.characters.reverse())
    }
    
}

```

masters3d:
the conflict seems to be here
``
    "house",
    "minesweeper",
    "ocr-numbers",
    "food-chain",
    "palindrome-products"
``

if you just keep this PR to 
    "palindrome-products"
it may help it.

masters3d:
This looks good. Let me know if I should merge.

robtimp:
@masters3d Sure, should be good to go now.

PR #117, opened 2016-02-15
Add Saddle Points exercise

robtimp:
@masters3d This is the last of the 'Suggested List of Problems post Swift 2.0'.

### DIFF
--- a/config.json
+++ b/config.json
@@ -65,7 +65,8 @@
     "ocr-numbers",
     "food-chain",
     "palindrome-products",
-    "pascals-triangle"
+    "pascals-triangle",
+    "saddle-points"
   ],
   "deprecated": [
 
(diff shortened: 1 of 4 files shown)


masters3d:
nice! good work. 

I did those because there seemed to be the most common amount other tracks. 

I'll be interested to see which of the rust and go problems we could implement in swift. Perhaps some of the a little more challenging problems in those tracks.

kytrinyx:
Check out the short section on porting exercises here: https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#porting-an-exercise-to-another-language-track

It describes how to get the full list of exercises that exist but are unimplemented in a track, as well as how to see which tracks have implemented it.

masters3d:
Cool! I still use http://synopsis.exercism.io/ for quick visualization but the api is nice!

On a side note: These tracks have not moved their problems so their links are the only ones working. 
xhaskel
xcoffeescript
xcpp
xlisp
xobjective-c
xperl5
xphp
xplsql
xracket
xscheme

kytrinyx:
Ah, yeah. I didn't realize synopsis is still up.

PR #129, opened 2016-03-31
Remove default multiples (issue #127)

robtimp:
Require all multiples to explicitly passed in, rather than using the default value of `[3,5]`, as specified in Issue #127.


robtimp:
@masters3d

masters3d:
Thanks! @robtimp

PR #130, opened 2016-04-01
switch position of the comma in JSON

masters3d:
This should help reduce the noise when reorganizing the order of problems
https://github.com/exercism/xswift/issues/56

Idealy I'd like to use the trailing comma but a "pre" comma should work the same
http://stackoverflow.com/questions/201782/can-you-use-a-trailing-comma-in-a-json-object


kytrinyx:
That is awkward and brilliant at the same time! I'm going to suggest that we do this across the board.

PR #132, opened 2016-04-03
dominoes: test singletons that can't be chained

### LINKED ISSUE #131: dominoes: Add test for input single domino (1, 2)
Some solutions mistakenly assume if the input has 1 domino it's always valid. This isn't always the case, so we should test for this by adding a test where the input has one domino `(1, 2)`. Adding in https://github.com/exercism/xrust/pull/96 , and it's just Rust and Swift that have the dominoes implemented.

petertseng:
Some solutions may automatically declare a singleton input to be valid
without checking it can actually be chained. This is incorrect behavior
and it is caught by this test.

Closes #131


masters3d:
Thanks!

PR #137, opened 2016-04-07
Update RnaTranscriptionTest.swift

### LINKED ISSUE #136: rna-transcription: don't transcribe both ways
I can't remember the history of this, but we ended up with a weird non-biological thing in the RNA transcription exercise, where some test suites also have tests for transcribing from RNA back to DNA. This makes no sense.

If this track does have tests for the reverse transcription, we should remove them, and also simplify the reference solution to match.

If this track doesn't have any tests for RNA->DNA transcription, then this issue can be closed.

See https://github.com/exercism/x-common/issues/148

masters3d:
closes #136

### DIFF
--- a/exercises/rna-transcription/RnaTranscriptionExample.swift
+++ b/exercises/rna-transcription/RnaTranscriptionExample.swift
@@ -4,7 +4,7 @@ struct Nucleotide {
     
     var complementOfDNA:String{ return transcribe(dnaToRna) }
     
-    private let value = ""
+    private let value
     
     init(_ nucleotide:String){
         self.value = nucleotide
--- a/exercises/rna-transcription/RnaTranscriptionExample.swift
+++ b/exercises/rna-transcription/RnaTranscriptionExample.swift
@@ -2,23 +2,21 @@
 
 struct Nucleotide {
     
-    var complementOfDNA:String{ return transcrive(DnaToRna) }
+    var complementOfDNA:String{ return transcribe(dnaToRna) }
     
-    private var value = ""
+    private let value = ""
     
     init(_ nucleotide:String){
-    self.value = nucleotide
-        
+        self.value = nucleotide
     }
     
-    private var DnaToRna:[Character:String] = [ "G": "C" ,  "C": "G" , "T": "A" ,  "A": "U" ]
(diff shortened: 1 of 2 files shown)


exercises/rna-transcription/RnaTranscriptionExample.swift
```diff
@@ -1,20 +1,17 @@
...
 
-
-
 struct Nucleotide {
     
     var complementOfDNA:String{ return transcrive(DnaToRna) }
-    var complementOfRNA:String{ return transcrive(RnaToDna) }
     
-    var value = ""
+    private var value = ""
```
robtimp: This can be a `let` since it's not changed after initialization.

exercises/rna-transcription/RnaTranscriptionExample.swift
```diff
@@ -1,20 +1,17 @@
...
+    
     init(_ nucleotide:String){
     self.value = nucleotide
         
     }
     
-    var DnaToRna:[Character:String] = [ "G": "C" ,  "C": "G" , "T": "A" ,  "A": "U" ]
-    var RnaToDna:[Character:String] = [ "C": "G" ,  "G": "C" , "A": "T" ,  "U": "A" ]
+    private var DnaToRna:[Character:String] = [ "G": "C" ,  "C": "G" , "T": "A" ,  "A": "U" ]
```
robtimp: Can also be a `let`.

exercises/rna-transcription/RnaTranscriptionExample.swift
```diff
@@ -1,20 +1,17 @@
...
 
     func transcrive(dict:[Character : String]) -> String{
```
robtimp: Typo. (transcrive -> transcribe)

robtimp:
@masters3d Commented on a couple minor things. Looks good otherwise.

exercises/rna-transcription/RnaTranscriptionExample.swift
masters3d: Thanks! I am doing the editing on the web interface :) 
"Immutable value 'self.value' may only be initialized once"

masters3d:
xcode does a lot for us:
"Type annotation missing in pattern"

masters3d:
@robtimp Thanks for taking a look!

PR #138, opened 2016-04-07
Update .travis.yml

masters3d:
Added `sudo: false`
https://github.com/exercism/xswift/pull/68

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ script:
   - xcodebuild test -project 'xcodeProject/xSwift.xcodeproj' -scheme 'xSwiftTests' -sdk macosx10.11
   - bin/fetch-configlet
   - bin/configlet .
+sudo: false


masters3d:
Ah this makes sence. Sudo false only works for linux builds. 
https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
image

kytrinyx:
oh, right. That does make sense.

masters3d:
We can leave it on when we move the XCTest to Linux

masters3d:
Do you know why there seems to be 2 instances of travis firing up?

image

kytrinyx:
Yeah, one is on the branch itself, and one is on the PR. It's been doing that for a while.

masters3d:
ah interesting, I never noticed. Thanks!

Issue #60, opened 2015-09-10
Make more (Error as return) throw-able problems for swift 2


masters3d:
- Replace Some optional returns for throws when it makes sense
- Prefer encapsulation instead of free functions. 
- Protocol usage when it makes sense
- Make extensions private
- Guard or if lets instead of `!`

masters3d:
This could be moved to a standard swift error system
https://github.com/exercism/xswift/tree/master/exercises/grains

PR #143, opened 2016-04-25
Add error handling to Grains exercise

robtimp:
Adds try/catch error handling to Grains exercise. (Partially addresses Issue #60.)

Also, improves `total` property in example by calculating it, rather than relying on the fact that it's the same as `UINT64_MAX`.

### DIFF
--- a/exercises/grains/GrainsExample.swift
+++ b/exercises/grains/GrainsExample.swift
@@ -16,16 +16,33 @@ func == (left: (result: UInt64, error: String?), right:(result: UInt64, error: S
 
 
 struct Grains {
-    static func square(num:Int)->(result: UInt64, error: String?){
-        if num < 1 || num > 64 {
-            return (result: 0, error: "Input[\(num)] invalid. Input should be between 1 and 64 (inclusive)")
-        } else {
-            let one:UInt64 = 1
-            let shift = UInt64(num - 1)
-            return  (result: one << shift, error:nil)
-        }}
+    enum Error: ErrorType {
+        case InputTooLow(String)
+        case InputTooHigh(String)
+    }
     
+    static func square(num: Int) throws -> UInt64 {
+        guard num >= 1 else {
+            let message = "Input[\(num)] invalid. Input should be between 1 and 64 (inclusive)"
+            throw Error.InputTooLow(message)
+        }
+        
+        guard num <= 64 else {
+            let message = "Input[\(num)] invalid. Input should be between 1 and 64 (inclusive)"
+            throw Error.InputTooHigh(message)
+        }
+       
+        let one: UInt64 = 1
+        let shift = UInt64(num - 1)
+        return one << shift
+    }
+        
     static var total:UInt64 {
-        return UINT64_MAX
+        let numbers = (1...64).map { $0 }
+        
+        return numbers.reduce(UInt64(0)) {
+            let squared = try! square($1)
+            return $0 + squared
+        }
     }
 }
--- a/exercises/grains/GrainsExample.swift
+++ b/exercises/grains/GrainsExample.swift
@@ -2,34 +2,23 @@ import Darwin
 
 
 
-func == (left: (result: UInt64, error: String?), right:(result: UInt64, error: String?) ) -> Bool
-{
-    if left.result == right.result && left.error == right.error
-    {
-        return true
-    }
-    else
-    {
-        return false
-    }
-}
-
 
 struct Grains {
-    enum Error: ErrorType {
-        case InputTooLow(String)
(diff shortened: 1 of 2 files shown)


robtimp:
@masters3d

exercises/grains/GrainsTest.swift
```diff
@@ -3,51 +3,90 @@ import XCTest
...
+        }
+        
     }
     
     func testInvalidInput2() {
-        XCTAssertTrue((result: 0, error: "Input[0] invalid. Input should be between 1 and 64 (inclusive)") == Grains.square(0))
+        var throwsInputTooLowError = false
+        
+        defer {
+            XCTAssertTrue(throwsInputTooLowError)
```
masters3d: This seems duplicate to me. Could you move this to the else block?

exercises/grains/GrainsTest.swift
```diff
@@ -3,51 +3,90 @@ import XCTest
...
+        }
+        
+        do {
+            let _ = try Grains.square(0)
+        } catch Grains.Error.InputTooLow(let message) {
+            throwsInputTooLowError = true
+            XCTAssertTrue(message == "Input[0] invalid. Input should be between 1 and 64 (inclusive)")
+        } catch {
```
masters3d: This is where you could put the defer code that fails.

exercises/grains/GrainsExample.swift
```diff
@@ -16,16 +16,33 @@ func == (left: (result: UInt64, error: String?), right:(result: UInt64, error: S
...
-    static func square(num:Int)->(result: UInt64, error: String?){
-        if num < 1 || num > 64 {
-            return (result: 0, error: "Input[\(num)] invalid. Input should be between 1 and 64 (inclusive)")
-        } else {
-            let one:UInt64 = 1
-            let shift = UInt64(num - 1)
-            return  (result: one << shift, error:nil)
-        }}
+    enum Error: ErrorType {
+        case InputTooLow(String)
```
masters3d: enum cases are now lower cased.

exercises/grains/GrainsTest.swift
```diff
@@ -3,51 +3,90 @@ import XCTest
...
-        XCTAssertTrue((result: 0, error: "Input[65] invalid. Input should be between 1 and 64 (inclusive)") == Grains.square(65))
+        var throwsInputTooHighError = false
+        
+        defer {
+            XCTAssertTrue(throwsInputTooHighError)
+        }
+        
+        do {
+            let _ = try Grains.square(65)
+        } catch Grains.Error.InputTooHigh(let message) {
```
masters3d: Are you able to say .inputTooHigh(let message ). This way people could place the enum where ever they want.

exercises/grains/GrainsTest.swift
```diff
@@ -3,51 +3,90 @@ import XCTest
...
+        
+        defer {
+            XCTAssertTrue(throwsInputTooLowError)
+        }
+        
+        do {
+            let _ = try Grains.square(-1)
+        } catch Grains.Error.InputTooLow(let message) {
+            throwsInputTooLowError = true
+            XCTAssertTrue(message == "Input[-1] invalid. Input should be between 1 and 64 (inclusive)")
```
masters3d: You should probably use xCTAsssertEquals here

exercises/grains/GrainsTest.swift
```diff
@@ -3,51 +3,90 @@ import XCTest
...
+        } catch {
+            return
+        }
     }
     
     func testSquare1() {
-        XCTAssertTrue((result: 1, error: nil) == Grains.square(1) )
+        XCTAssertTrue(try! Grains.square(1) == 1)
     }
```
masters3d: Should probably move this to assert equals too.

exercises/grains/GrainsTest.swift
robtimp: This block would only be executed if an error (other than `.inputTooLow`) is thrown, though. It would miss the path of the try actually succeeding. That was the reason for the `defer`.

exercises/grains/GrainsTest.swift
robtimp: Unfortunately, it's not able to infer the error types that way: it only works if I fully qualify the name.

exercises/grains/GrainsTest.swift
robtimp: See comment below: there's three possible paths that need to be covered.

robtimp:
@masters3d Updated

masters3d:
Thank you!

PR #147, opened 2016-06-04
Remove warnings for vars not modified

jonallured:
This particular test was using `var`s but then not mutating them, which
was causing Xcode to issue warnings.

### DIFF
--- a/exercises/binary-search-tree/BinarySearchTreeTest.swift
+++ b/exercises/binary-search-tree/BinarySearchTreeTest.swift
@@ -9,28 +9,28 @@ class BinarySearchTreeTest: XCTestCase {
     }
     
     func testInsertingLess() {
-        var four = BinarySearchTree(4)
+        let four = BinarySearchTree(4)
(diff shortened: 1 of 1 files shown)


masters3d:
We left the var here because there are solutions that mutate by using an indirect enum. I don't want to disallow solution using value types.

jonallured:
Cool, what would you think about ignoring the file for warnings then, something like:

http://stackoverflow.com/questions/6921884/in-xcode-how-to-suppress-all-warnings-in-specific-source-files

masters3d:
that sounds good. I don't think it works with swift? haven't tried it. 
image

jonallured:
Darn, guess it was an Obj-C file I did this for. :\

http://stackoverflow.com/questions/31540446/how-to-silence-a-warning-in-swift-2-0/32861678#32861678

masters3d:
I guess we could implement that particular problem using enums to get rid of the warning in our Xcode project.

PR #148, opened 2016-06-04
Remove extra project

jonallured:
Only the hello-world exercise had a corresponding project file, which seemed weird, so this commit gets it back in line with the other exercises.


masters3d:
Sorry this is meant to help very beginners get started.  https://github.com/exercism/exercism.io/issues/2401#issuecomment-104283971

kytrinyx:
Also: Thank you so much for getting involved! <3 It helps so much when people have opinions and thoughts (even when we close things because there's history that isn't necessarily well-documented). :heart:

jonallured:
Thanks @kytrinyx and thanks @masters3d for the feedback!! :D

I guess I still think it's weird to have the first exercise different than the rest without any explanation, but maybe something can be done there. Regardless, what would you think about a PR with just f2f104115? Unless I'm missing something, that scheme doesn't work and it's necessary.

Thanks!!

masters3d:
yeah that schema can go away. In fact I am hoping that we can move away from Xcode projects once we have the swift package manager and there is better support for it on windows.

kytrinyx:
How about adding some comments to the first exercise that explains that we've bootstrapped things to help people get started, and that we won't be doing that on the next exercises, but if they want to they can always look at this first exercise for inspiration if they get stuck on boilerplate/project setup?

PR #149, opened 2016-06-04
Normalize test class and filenames

jonallured:
Seemed like `SomethingTest` and SomethingTest.swift were the most popular, so I went with those. While I was at it, I also sorted the exercises in the Project navigator so that it was easier to find things.


masters3d:
This one looks good! Thank you

Issue #10, opened 2014-08-17
Recommend a linter for Swift?


kytrinyx:
This may or may not be appropriate in this language. See exercism/docs#54 for details.

masters3d:
There is an experimental one

masters3d:
We probably wont have a real linter for some time.

yurrriq:
:+1: for SwiftLint!

jonallured:
I've been using SwiftLint on a production project and we've really liked it. Also, houndci.com supports Swift. What's important to me is having some sort of consistency, which ruleset is used less so.

What can I do to move this forward?

masters3d:
hound looks interesting. https://houndci.com/configuration#swift 
@kytrinyx what do you think?

masters3d:
This would be something that would fit perfectly when we move to swift 3 in a couple of months.

kytrinyx:
I haven't written any swift, so I don't have a basis for comparison.

@marianaIAm have you by any chance written enough Swift in production to have experiential knowledge of linters and style rules?

yurrriq:
It looks like Hound CI wraps SwiftLint. I've found SwiftLint to be great over the last n months and would definitely recommend it. Perhaps the real question is: what's a good config to recommend?

masters3d:
I think the default rules are fine except for these:
```
disabled_rules: # rule identifiers to exclude from running
  - force_try
  - function_body_length
  - line_length
  - variable_name
```

The reason is that some of the tests have really long lines while others use force try to keep the code simple for the tests. Variable names rule is really opinionated as it won't even allow short variables like `i`

I am going to turn on the swiftlinter as part of Travis when we update to swift 3.0
https://github.com/exercism/xswift/pull/155

jonallured:
@masters3d agreed!

The only quibble I'd have is the variable names one: I think short var names are an anti-pattern, but make not worth having the linter complain?

masters3d:
I guess instead of disabling `variable_name` we could downgrade it to a warning instead of an error. 

```
variable_name: 
    min_length: 
        warning: 1
        error: 0

force_try: warning
    
disabled_rules: # rule identifiers to exclude from running
  - function_body_length
  - line_length
```

masters3d:
We are adopting SwiftLint for swift3.

PR #150, opened 2016-06-04
Whitespace Cleanup

jonallured:
I know, I know, whitespace - who cares? It was just really bothering me that there was so much inconsistency, so I did my best to resist frivolous changes and stuck to mostly running:

Editor > Structure > Re-Indent

With 4 spaces as my indent. It's about as close as we can get to `gofmt` and seemed reasonable. Most of the files I saw matched this format anyway, many just had a mistake here or there.

Then I just went through and removed some bonus newlines. Ahhh, much better!! :sunglasses:

### DIFF
--- a/exercises/accumulate/AccumulateExample.swift
+++ b/exercises/accumulate/AccumulateExample.swift
@@ -1,6 +1,4 @@
-
 // Foundation not needed
-
 
 extension Array {
     func accumulate<S>(yield: (Element) -> S) -> [S] {
--- a/exercises/accumulate/AccumulateTest.swift
+++ b/exercises/accumulate/AccumulateTest.swift
@@ -1,12 +1,8 @@
-
 import XCTest
 
-
-
 private extension String {
-    
-    var length: Int {return self.characters.count}
-    
+    var length: Int { return self.characters.count }
+
     func reverse() -> String {
         var result:String = ""
         for char in self.characters {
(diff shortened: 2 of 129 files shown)


masters3d:
Oh I know!  I wish we had a standard swift format tool like gofmt but we don't have that yet. 
https://github.com/exercism/xswift/issues/10

I am not sure what the stand is for exercism as a project on whitespace only PRs but this being a swift language track I am going to close this for now until we have an official swift linter. 

References:
https://github.com/apple/swift/pull/275
https://github.com/apple/swift/pull/963

kytrinyx:
I think that if we are internally inconsistent, we could probably choose some sort of rule-set just for the repo, and be consistent with that (in which case whitespace-only PRs would totally be a good thing). I think the question is--are we inconsistent? If so, what sort of consistency is important? If we make it consistent, is there some sort of mini-linting that we could do to make it easy to keep things consistent (and also something that we could run on CI)?

masters3d:
I would love a CI solution that makes a lint commit. . I have heard of something like that but I am not sure how to get it set up. I know travis can check lint rules but I don't know about actually making a "Lint" commit.  https://www.bignerdranch.com/blog/continuous-linting/

jonallured:
Makes sense, I'll try to make progress on #10 instead. Thanks!! :D

PR #151, opened 2016-06-05
Remove unnecessary scheme

jonallured:
As discussed in #148, this scheme doesn't work and is not necessary. <3


masters3d:
Thanks!

PR #154, opened 2016-06-16
Copy track icon into language track repository

martinsvalin:
See exercism/exercism.io#2925 for more details.


masters3d:
Thanks!

PR #155, opened 2016-06-17
Swift 3 Preview 1 Change

### LINKED ISSUE #10: Recommend a linter for Swift?

masters3d:
I made these changes to swift 3 but I'd like to wait and leave this PR open for a couple of weeks. There are a couple of changes to swift 3 that haven't landed but this will give us a good idea on the changes. 

I think the most interesting changes have to do with thew new collection indexing model. 

closes #10


kytrinyx:
@masters3d What kind of review would help you here? I can't provide it, but I could help see if I can find someone who can help.

robtimp:
@masters3d @kytrinyx I’ve been reviewing it as changes are added, but it shouldn’t be merged into `master` until the official release of Swift 3. 

It may be helpful to create a separate Swift 3 branch until then.

masters3d:
I am not sure how to notify people who are currently doing the exercises that we are changing versions. Last time we moved to swift 2 I added a comment to the test file. In swift 3 I am able to require a minimum version as a compiler directive but it maybe confuse beginners. 

I think what I am missing is to have somebody take a quick glance and help me decide if we should go ahead with swift 3 early. I'm eager to transition and the tools are here. 


This PR is 
- upgrading code to Swift 3
- adding support to SwiftLint
- updating documentation 

The code upgrade was fairly mechanical. SwiftLint caused lots of noise because I ran autocorrect on the repo to get Travis to pass.

masters3d:
@robtimp I thought about a separate branch but I really don't want to keep separate branches for the next 3 months. 

I also thought about maintaining support for swift 2 along with swift 3 but it became way too coversome. 

Xcode 8 is available now (is not GM tough) so the tools are there not to mention Ubuntu support (which I want to support a little later. )

kytrinyx:
> I am not sure how to notify people who are currently doing the exercises that we are changing versions.

We have "system" notifications which would show up in people's notifications page. We could write a script to just notify people who are doing swift.

masters3d:
closing in favor of a separate swift3 branch https://github.com/exercism/xswift/pull/160

PR #157, opened 2016-07-12
Update TESTS.md

### LINKED ISSUE #156: Initial Setup of Swift projects unclear
There are several hurdles with beginning the Swift track would benefit from some attention. I am very new to exercism and somewhat new to programming, and I respect how much work it is to create these assets. These observations are offered as constructive criticism, and I do not mean to disparage the work already performed.

1. The "Running the Tests" instructions are based on the bob exercise, rather than the hello-world exercise.

2. The source files retrieved by the fetch operation for the hello-world exercise are not helpful as they do not match other exercises. For example, most exercises fetch two source documents, tests and a readme. Hello World contains an Xcode project and directory. And worse, the fetched Xcode project does not build as a standalone project.

_These two issues created a lot of false inferences on my part as I tried to get up to speed_. I ultimately worked out how exercism worked with Swift by finding the backing GitHub repositories for bob and hello-world, and comparing them to the instructions.

Recommendations

The "Running the Tests" instructions for Swift are strictly correct, although I still found them confusing until I figured out the basic workflow. (Again hello-world sent me off in the wrong direction initially.)

A few specific notes that might help make the running test instructions mode complete. 

a. Note that its better to use iOS Single View application even if you're ultimately going to work on OS X apps. 

b. I think asking newer developers to create a class file outside Xcode's normal default is a tall order, and it would be better to leave it in place. Then demonstrate how to submit it by dragging and dropping the class file from Xcode into the terminal window.

c. describe briefly difference between unit tests and compiling the project

d. add several additional images.

Expanding on the Running the Tests instructions to explain both how to create a new project and how to upload the final submission is something I would like to contribute to, but I'm not sure how. I love the whole exercism concept and would love to contribute back to it.

masters3d:
closes #156

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -1,5 +1,7 @@
 ### Creating the project in Xcode
 
+Please see [Apple's guide on XCTest](https://developer.apple.com/library/tvos/documentation/DeveloperTools/Conceptual/testing_with_xcode/chapters/02-quick_start.html#//apple_ref/doc/uid/TP40014132-CH2-SW1)  
+
 These steps will use the `bob` exercise as an example. If you are not working on `bob`, replace `bob` with your exercise name in the steps below.
 
 * Select iOS-->Application and then Single View Application and click Next.


robtimp:
@masters3d We'll probably have to update periodically, since Apple has a habit of changing documentation links.

masters3d:
yeah. There is probably better links out there too. I guess we should do that for objc track too at some point.

bobdel:
I found parts one and two of this tutorial was all that I needed to understand exercism's approach.

https://medium.com/@ynzc/getting-started-with-tdd-in-swift-2fab3e07204b#.qd08pnfpz

kytrinyx:
> I found parts one and two of this tutorial was all that I needed to understand exercism's approach.

That's excellent! What do you all think of pointing to that from TESTS.md?

PR #162, opened 2016-07-19
Fix warnings and clean up

robtimp:
* Fix all warnings for Xcode 8 beta 3
* Simplify GradeSchoolTest.swift

### DIFF
--- a/exercises/grade-school/GradeSchoolTest.swift
+++ b/exercises/grade-school/GradeSchoolTest.swift
@@ -2,44 +2,22 @@
     import XCTest
 #endif
 
-private extension XCTestCase {
-
-    // Workaround unit test liminitaiton on Optional Arrays/Sets
-    private func sameCollection<C: Collection where C.Iterator.Element == String>(_ result: C?, _ expected: [String]?) -> Bool {
-        guard let result = result, expected = expected where similar(result, expected) else { return false }
-        for (index, student) in result.enumerated() {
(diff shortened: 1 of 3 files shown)


robtimp:
@masters3d

exercises/grade-school/GradeSchoolTest.swift
```diff
@@ -2,84 +2,58 @@
...
-        return result.count.toIntMax() == IntMax(expected.count)
-    }
-
-    func XCTAssertEqualCollection (_ collectionS: Set<String>?, _ collectionA: [String]? ) {
-        XCTAssert(sameCollection(collectionS, collectionA))
-    }
-
-    func XCTAssertEqualCollection (_ collectionA1: [String]?, _ collectionA2: [String]? ) {
-        XCTAssert(sameCollection(collectionA1, collectionA2))
-    }
```
masters3d: So there is an issue here with people trying to use a set instead of an array or the other way around that's why we added this. Do you think we can support both with out it?

masters3d:
Do you think we should force a set? I'm actually okay with that so the code gets simple and people get to use a set.

masters3d:
Now we have Generic typealias in swift3 so we could possibly do that for the array/set

https://github.com/exercism/xswift/pull/83

robtimp:
@masters3d I think ideally we should support `Array` or `Set`. It _seems_ like there ought to be some kind of more elegant solution, but I haven’t been able to find one so far.

PR #169, opened 2016-08-17
autocorrect --format now runs on compile

masters3d:
- fixed all warnings
- turn on autocorrect --format on compile


kytrinyx:
This looks good to me.

masters3d:
Thanks!

PR #174, opened 2016-09-03
Add Run Length Encoding exercise

robtimp:
@masters3d

### DIFF
--- a/exercises/run-length-encoding/RunLengthEncodingTest.swift
+++ b/exercises/run-length-encoding/RunLengthEncodingTest.swift
@@ -1,0 +1,34 @@
+#if swift(>=3.0)
+    import XCTest
+#endif
+
+class RunLengthEncodingTest: XCTestCase {
+    
+    func testEncodeSimple() {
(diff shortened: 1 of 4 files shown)


masters3d:
travis probably rolled back to 10.11
could you change that back to 10.11 from 10.12 in the travis config?

masters3d:
I think I fixed it.

exercises/run-length-encoding/RunLengthEncodingTest.swift
```diff
@@ -0,0 +1,32 @@
+import XCTest
```
masters3d: Can you add the swift 3 check here?

masters3d: I think there are going to have a better flag later but for now I think having #if swift check would work.

masters3d:
Thanks! we are getting close to release the swift3 version. I think as soon as xcode goes GM, we pull the trigger.

robtimp:
Great!

PR #185, opened 2016-09-09
Update config.json to new specification

Roshanjossey:
This addresses issue #170
https://github.com/exercism/xswift/issues/170

### DIFF
--- a/config.json
+++ b/config.json
@@ -70,6 +70,398 @@
    ,"bowling"
    ,"run-length-encoding"
   ],
+  "exercises": [
+    {
+      "slug": "hello-world",
+      "difficulty": 1,
+      "topics": [
+      ]
+    },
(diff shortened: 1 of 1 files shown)


masters3d:
Thanks!

In all honesty I was planning to just steal Fsharp's and iterate from there. https://github.com/exercism/xfsharp/blob/master/config.json

Do you mind updating using fsharp's keywords and difficulties ?

Roshanjossey:
@masters3d: I was thinking about making this change in multiple steps as described in the issue

> 1. Add the exercises key with the array of objects, where difficulty is 1 and topics is empty.

I'd prefer to send another PR after adding difficulty and topics from Fsharp stream

masters3d:
Okay. Thanks!

PR #195, opened 2016-10-20
Add Pangram exercise

robtimp:
Includes Foundation in order to strip diacritics from String. The last exercise could be removed if this is not desirable.

### DIFF
--- a/exercises/pangram/PangramExample.swift
+++ b/exercises/pangram/PangramExample.swift
@@ -1,0 +1,17 @@
+import Foundation
+// Needed to strip diacritics
+
+struct Pangram {
+
+    static func isPangram(_ text: String) -> Bool {
+        let lowercasedText = text.folding(options: .diacriticInsensitive, locale: .current).lowercased()
+
+        for letter in "abcdefghijklmnopqrstuvwxyz".characters {
+            if lowercasedText.contains(String(letter)) == false {
+                return false
+            }
+        }
+
+        return true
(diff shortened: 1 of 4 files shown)


robtimp:
@masters3d

exercises/pangram/PangramExample.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+// Needed to strip diacritics
```
masters3d: I am good with foundation now that it is part of the swift open source.

exercises/pangram/PangramExample.swift
```diff
@@ -0,0 +1,17 @@
...
+
+struct Pangram {
+
+    static func isPangram(_ text: String) -> Bool {
```
masters3d: What do you think of making this an extension on String

```swift
var isPangram:Bool {...}

robtimp: I guess my personal preference is not to do this, at least for the purposes of this project. It seems cleaner to me to write a separate Class or Struct per exercise, rather than extending the built-in types.

masters3d: Ah interesting, I think it is cleaner to extend builtin types. This is a simple exercise so I agree on putting on struct but it would be cool to figure out a way to require the test to use protocols. 

```swift
protocol Pangram {
   var isPangram:Bool
}

PR #198, opened 2016-11-07
Clean up PalindromeProducts

### LINKED ISSUE #188: PalindromeProducts Random Test Failures.
I suspect this has something to do with multi-thread code causing a race condition.

```
Test Suite 'PalindromeProductsTest' started at 2016-09-09 23:11:08.621
Test Case '-[xSwiftTests.PalindromeProductsTest testLargestPalindromeFromSingleDigitFactors]' started.
Test Case '-[xSwiftTests.PalindromeProductsTest testLargestPalindromeFromSingleDigitFactors]' passed (0.004 seconds).
Test Case '-[xSwiftTests.PalindromeProductsTest testLargestPalindromeFromTripleDigitFactors]' started.
Test Case '-[xSwiftTests.PalindromeProductsTest testLargestPalindromeFromTripleDigitFactors]' passed (2.329 seconds).
Test Case '-[xSwiftTests.PalindromeProductsTest testSmallestPalindromeFromDoubleDigitFactors]' started.
Test Case '-[xSwiftTests.PalindromeProductsTest testSmallestPalindromeFromDoubleDigitFactors]' passed (0.022 seconds).
Test Case '-[xSwiftTests.PalindromeProductsTest testSmallestPalindromeFromTripleDigitFactors]' started.
Test Case '-[xSwiftTests.PalindromeProductsTest testSmallestPalindromeFromTripleDigitFactors]' passed (2.387 seconds).
Test Suite 'PalindromeProductsTest' failed at 2016-09-09 23:11:13.375.
     Executed 5 tests, with 1 failure (0 unexpected) in 4.742 (4.754) seconds
```

robtimp:
Resolves #188 

* (Hopefully) fixes multi-threading crash in PalindromeProducts tests.
* Simplify the method for reversing a String
* Minor clean-up

### DIFF
--- a/exercises/palindrome-products/PalindromeProductsExample.swift
+++ b/exercises/palindrome-products/PalindromeProductsExample.swift
@@ -2,14 +2,10 @@ import Foundation
 
 private extension String {
 
-    var length: Int {return self.characters.count}
+    var length: Int { return self.characters.count }
 
     func reverse() -> String {
-        var result: String = ""
-        for char in self.characters {
-            result = "\(char)\(result)"
-        }
-        return result
+        return characters.reversed().map { String($0) }.joined()
     }
 }
 
@@ -19,8 +15,8 @@ struct PalindromeProducts {
     private let maxFactor: Int
     private let minFactor: Int
 
-    var largest: Palindrome {return calculate(.max)}
-    var smallest: Palindrome {return calculate(.min)}
+    var largest: Palindrome { return calculate(.max) }
+    var smallest: Palindrome { return calculate(.min) }
 
     init(maxFactor: Int, minFactor: Int = 1) {
         self.maxFactor = maxFactor
@@ -31,7 +27,6 @@ struct PalindromeProducts {
     private func calculate(_ upTo: Mode) -> Palindrome {
 
         let rangeOuter = minFactor...maxFactor
-        var multiplications = [Palindrome]()
 
         //Multithreaded code
         var results = [[Palindrome]](repeating: [Palindrome](), count: rangeOuter.count)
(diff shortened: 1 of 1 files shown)


robtimp:
@masters3d I’ve run the test many times without seeing the crash again, but it’s pretty difficult to prove it could _never_ happen, so let me know if you do still see it.

exercises/palindrome-products/PalindromeProductsExample.swift
```diff
@@ -31,7 +27,6 @@ struct PalindromeProducts {
     private func calculate(_ upTo: Mode) -> Palindrome {
 
         let rangeOuter = minFactor...maxFactor
-        var multiplications = [Palindrome]()
 
         //Multithreaded code
         var results = [[Palindrome]](repeating: [Palindrome](), count: rangeOuter.count)
```
masters3d: I believe the issue with the failing test has to do with the ``results`` array. 
http://stackoverflow.com/questions/28191079/create-thread-safe-array-in-swift/28976644#28976644
If the array were to be written to from two threads at the same time, it will have problems and this is why it doesn't happen often. 
https://www.reddit.com/r/swift/comments/4myh3w/is_swift_array_thread_safe_can_swift_array_face/

we need to lock the array when writing to it to prevent other threads from writing to it at the same time. 
https://www.bignerdranch.com/blog/locking-in-swift-helping-protect-me-from-myself/

exercises/palindrome-products/PalindromeProductsExample.swift
```diff
@@ -51,17 +46,17 @@ struct PalindromeProducts {
             }
             results[advanceByIndex] = multiplicationsTemp
```
masters3d: I think we can fix it for sure by creating a "write thread", so in line 47 we dispatch a write queue that we use to write to the results.  

http://stackoverflow.com/questions/24045895/what-is-the-swift-equivalent-to-objective-cs-synchronized

masters3d:
I think we can avoid the current data race by creating a read/write thread for the  `results` array. See comments.

masters3d:
More context. https://twitter.com/cocoaphony/status/793885415237361664

PR #206, opened 2016-11-25
robot update

masters3d:


kytrinyx:
👍 

I also wrote a bit about _why_ we want a summary in the pull request description. Should we link to that? http://tinyletter.com/exercism/letters/exercism-pull-requests

masters3d:
Oh good point.

PR #209, opened 2016-12-05
Add Transpose exercise

robtimp:
* Add Transpose exercise

In contrast with the Ruby version, the Swift version operates on arrays of `String`s, since the syntax for multi-line strings is rather cumbersome.

### DIFF
--- a/exercises/transpose/TransposeExample.swift
+++ b/exercises/transpose/TransposeExample.swift
@@ -1,0 +1,35 @@
+struct Transpose {
+    static func transpose(_ input: [String]) -> [String] {
+        let maxLineLength = input.map { $0.characters.count }.max()
+
+        guard let maxLength = maxLineLength else {
+            return []
+        }
+
+        var result = [String](repeatElement("", count: maxLength))
+
+        for i in 0..<maxLength {
+            for line in input {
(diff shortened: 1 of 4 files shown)


robtimp:
@masters3d

masters3d:
Nice!

exercises/transpose/TransposeExample.swift
```diff
@@ -0,0 +1,35 @@
...
+                    let end = line.index(start, offsetBy: 1, limitedBy: line.endIndex) {
+                    let character = line[start..<end]
+                    result[i].append(character)
+                } else {
+                    result[i].append(" ")
+                }
+            }
+        }
+
+        return result.map { stripTrailingWhitespace($0) }
```
regnerjr: Great use of map! I just wanted to point out that map can also take a function as a parameter. Instead of the trailing closure which is explicitly calling the function you could instead write. 
```
return result.map(stripTrailingWhitespace)
```
Which is a little cleaner.

masters3d: good call. That would be an awesome rule to add to Swiftlint. You should suggest it.

masters3d: https://github.com/realm/SwiftLint/issues/929

PR #210, opened 2016-12-05
Fix stylistic warnings

robtimp:
* Fix style warnings from Danger


robtimp:
@masters3d

exercises/crypto-square/CryptoSquareExample.swift
```diff
@@ -64,12 +64,13 @@ struct Crypto {
...
 
     var normalizeCiphertext: String {
-
         let sizeNormal: Int = (ciphertext.characters.count == self.size * self.size ) ? getSquareSize(self.ciphertext) : getSquareSize(self.ciphertext, floorNoCeling: true)
 
         return segmentSorter(ciphertext, spacing: sizeNormal).joined(separator: " ")
     }
-    var plaintextSegments: [String] { get { return segmentSorter(self.normalizePlaintext, spacing: self.size)}}
+    var plaintextSegments: [String] {
+        return segmentSorter(self.normalizePlaintext, spacing: self.size)
```
masters3d: I prefer to only include self when required by the compiler.
In this project we should not required but we don't have a way to enforce that. 
There is a rule being worked on that would required them. 
https://github.com/realm/SwiftLint/pull/829
I wonder if there is an easy way to create this rule with swiftlint.

PR #212, opened 2016-12-06
Update INSTALLATION.md

bobdel:
This minor change adds a link to the Mac App Store.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,6 +1,9 @@
 ## Installing Swift
 
-In order to use Swift, you must be running Xcode version 8.0 or greater which is available  at [Apple's developer center](https://developer.apple.com/xcode/downloads/).
+In order to use Swift, you must be running Xcode version 8.0 or greater. 
 
+The current release version is available at the [Mac App Store](https://itunes.apple.com/us/app/xcode/id497799835?mt=12). 
+
+Additional versions of Xcode (including the most recent beta) are available at [Apple's developer center](https://developer.apple.com/xcode/downloads/).
 
 All exercises tested with Xcode 8.0 using Swift 3


masters3d:
Thank you!

bobdel:
I'm working on a longer version of the setup document, and would like to include images. I'm still pretty new to Github, so I'd welcome advice no matter how basic.

As I understand it, I should add an image to a message like this, then once its uploaded grab its URL. Is that correct?

masters3d:
You are welcome to upload the images anywhere where they are known to have long retention of direct links. (Flickr, GitHub etc) I personally just like using GitHub which it is really a hack. 
https://github.com/exercism/xswift/pull/204
Perhaps the proper way would be to add the images to a folder and reference them that way. 
I'll leave that up to you. We have a logo images folder. 
Thanks again!

masters3d:
https://github.com/exercism/xswift/tree/master/img
You are welcome to use this folder for documentation images.

kytrinyx:
Yeah, for contributing documentation I think adding them to `img/` is the right thing.

bobdel:
Since I don't have push access to that directory, how would I go about adding images to that directory?

masters3d:
https://github.com/exercism/xswift/pull/214#issuecomment-265614616

kytrinyx:
I am stranded at an airport, so my vacation is a bit delayed.

I was confused when I posted `img/`, I actually meant `docs/img`. But even this is broken. I've opened a PR so we can fix it. https://github.com/exercism/trackler/pull/15

kytrinyx:
Wait. I'm confused about being confused.

Images in documentation about **contributing to the language track** should go in the top level `img/` directory.

Images in documentation **meant to be rendered on the exercism.io/languages/swift pages** (about how to get started solving the exercises in Swift) should go in `docs/img` (which you can create if it doesn't exist).

masters3d:
Ah that clears it up. Thanks for the help.

PR #221, opened 2016-12-13
Solution for the sublist exercise

AlwynC:
Implemented the solution for the sublist exercise mentioned in https://github.com/exercism/x-common/blob/master/exercises/sublist/description.md.


exercises/sub-list/SublistExample.swift
```diff
@@ -0,0 +1,71 @@
+import Foundation
+
+struct SubList {
+
+    var firstList: [Int]
+    var secondList: [Int]
+
+    func checkSubList(listOne: [Int], listTwo: [Int]) -> String {
```
masters3d: this should be a static funcion

exercises/sub-list/SublistExample.swift
```diff
@@ -0,0 +1,71 @@
+import Foundation
+
+struct SubList {
+
+    var firstList: [Int]
```
masters3d: dont need these if the below funcion is static

exercises/sub-list/SublistExample.swift
```diff
@@ -0,0 +1,71 @@
...
+    var secondList: [Int]
+
+    func checkSubList(listOne: [Int], listTwo: [Int]) -> String {
+
+        if  listOne == listTwo {
+
+            return "equal"
```
masters3d: return should be an enum

exercises/sub-list/SublistExample.swift
```diff
@@ -0,0 +1,71 @@
...
+                    count+=1
+                }
+
+                if smallerList == listTwo {
+
+                    return "superlist"
+
+                } else if smallerList == listOne {
+
+                    return "sublist"
```
masters3d: enum .sublist

xcodeProject/xSwift.xcodeproj/project.pbxproj
```diff
@@ -1176,35 +1156,18 @@
...
-				},
-			);
 			projectRoot = "";
 			targets = (
 				9D3155D11BA12FB1009A3991 /* xSwiftTests */,
 			);
 		};
 /* End PBXProject section */
 
-/* Begin PBXReferenceProxy section */
```
masters3d: thanks fixing this!

exercises/hello-world/helloWorldTest.swift
```diff
@@ -1,6 +1,6 @@
 //
```
masters3d: do u mind deleting lines 1-4?

exercises/sub-list/SubListTest.swift
```diff
@@ -0,0 +1,128 @@
+import XCTest
+
+class SubListTest: XCTestCase {
+
+    func testEmptyLists() {
+
+        let sublists = SubList(firstList: [], secondList: [])
+        let expected = "equal"
+        XCTAssertEqual(expected, sublists.checkSubList(listOne: sublists.firstList, listTwo: sublists.secondList))
```
masters3d: inline expected value here as `.equal`

exercises/sub-list/SublistExample.swift
masters3d: `func classifier(listOne:[Int], listTwo:[Int]) -> SublistKind`

AlwynC:
@masters3d I've made the suggested changes in my latest commit.

exercises/sub-list/SublistExample.swift
```diff
@@ -1,71 +1,72 @@
 import Foundation
 
-struct SubList {
+enum SublistKind: String {
```
robtimp: The string value of the enum isn't used anywhere. Can be changed to just `enum SublistKind {`.

exercises/sub-list/SublistExample.swift
```diff
@@ -0,0 +1,72 @@
+import Foundation
+
+enum SublistKind: String {
```
robtimp: Please remove the raw value, since it's unused.

exercises/sub-list/SubListTest.swift
```diff
@@ -0,0 +1,95 @@
...
+class SubListTest: XCTestCase {
+
+    func testEmptyLists() {
+
+        XCTAssertEqual(SublistKind.equal, classifier(listOne: [], listTwo: []) )
+    }
+
+    func testEmptyListWithinNonEmptyList() {
+
+        XCTAssertEqual(SublistKind.sublist, classifier(listOne: [], listTwo: [1, 2, 3]))
```
masters3d: You can also drop the first part of the enum `XCTAssertEqual(.sublist, ...`
This is up to you. 
I like leaving out the first part because it doesn't force the implementor to use the same name for their enum.

exercises/sub-list/SublistExample.swift
AlwynC: Yes the raw value is unused,but I added it since the expected values for each test case shown in  
https://github.com/exercism/x-common/blob/master/exercises/sublist/canonical-data.json are given as strings so I thought we could extract the raw values if required.Anyways,I've removed the String raw value for SublistKind in my latest commit.

exercises/sub-list/SubListTest.swift
AlwynC: I generally add the enum name explicitly for accessing the enum values in the XCTAssert statements shown above for my clarity.However I've removed them in my latest commit.

kytrinyx:
The `sublist` exercise needs to be named `sublist`, not `sub-list`. I'm fixing this so that I can deploy.

AlwynC:
@kytrinyx Thanks for #222 .

PR #222, opened 2016-12-15
Fix name of sublist exercise

kytrinyx:
The sublist exercise is named sublist, not sub-list. The wrong name breaks the
fetch for people, so while this fix might not actually work correctly, it will
let us deploy.

Please check that this exercise still works and fix it in a separate PR if not.


kytrinyx:
FYI @masters3d @robtimp

robtimp:
@kytrinyx Verified. Thanks for catching this!

masters3d:
@kytrinyx sorry to drag you out of vacation! Thanks for fixing!

A discussion for a later time, how can we prevent this from happening again?

kytrinyx:
> sorry to drag you out of vacation! Thanks for fixing!

I was stranded at an airport, so it was no skin off my vacation :)

> A discussion for a later time, how can we prevent this from happening again?

The `configlet` tool should probably have an option to verify the exercise names. It could hit the x-api to get the full list. I opened an issue for it: https://github.com/exercism/configlet/issues/10

I finally made it out of the airports and airplanes, so I'll be disappearing into my vacation sometime today.

PR #228, opened 2016-12-19
Solution for flatten array exercise

AlwynC:
Implemented the flatten array exercise mentioned at https://github.com/exercism/x-common/blob/master/exercises/flatten-array/description.md .


exercises/flatten-array/FlattenArrayExample.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+func flattenArray(list: [Any?]) -> [Any] {
```
masters3d: This really should be a generic function.

exercises/flatten-array/FlattenArrayTest.swift
```diff
@@ -0,0 +1,42 @@
...
+        let intResult = result as? [Int]
+        XCTAssertEqual([0, 2, 2, 3, 8, 100, -2], intResult!)
+    }
+
+    func testFlattenForAllNullDeepNestedList() {
+
+        let nilValue: Any? = nil
+        let result = flattenArray(list: [nilValue, [[[nilValue]]], nilValue, nilValue, [[nilValue, nilValue], nilValue], nilValue])
+        let intResult = result as? [Int]
+        XCTAssertEqual([], intResult!)
```
masters3d: This unwrap here is odd.

masters3d: I would avoid force unwrapping

exercises/flatten-array/FlattenArrayTest.swift
```diff
@@ -0,0 +1,42 @@
+import XCTest
+
+class FlattenArrayTest: XCTestCase {
+
+    func testFlattenIntegerArray() {
+
+        let result = flattenArray(list: [1, [2, 3, 4, 5, 6, 7], 8])
+        let intResult = result as? [Int]
```
masters3d: This casting would not be nessesary if the functions was a generic constrained to the Element type inside the incoming array.

exercises/flatten-array/FlattenArrayExample.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    func extractArrayElements(array: [Any?]) {
+
+        for element in array {
+
+            if let j = element {
+
+                let anyObjectArray = j as? [Any?]
+
+                if let forceUnwrappedArray = anyObjectArray {
```
masters3d: This name here is misleading. force unwrapping only happens when we use the ! operator.

exercises/flatten-array/FlattenArrayExample.swift
masters3d: This is a actually a difficult problem to solve well while keeping strong typing. I would recommend constraining to  Int Array and 2 levels of nesting arrays.

masters3d: Specifically It would require Conditional Conformances which are not yet implemented 
https://github.com/apple/swift-evolution/blob/master/proposals/0143-conditional-conformances.md

masters3d:
This particular excercise is depricated in other tracks like f-shap. I am not sure if there is a list of depricated excersices but this one is not really interesting in strongly typed languages like swift unless we make it replicate the functionality of .joined() as an advance excercise. 

I am sorry but as it stands I don't think we should accept this particular excercise. 

What do you think about closing this for now but 
if you feel that once swift 4 comes out with conditional conformances this excercise could be written in such a way that it is strongly typed then you are welcome to open this up again. 

I don't know who decides when an excersice is depricated.

AlwynC:
@masters3d Yes this is a difficult problem to solve by trying to keep up with strong typing.I've just solved this exercise without using generics for constraining the type of array elements.Please discuss this with other members of your team if this exercise is to be deprecated or not.If this exercise is to be deprecated,then I'll close this pull request and re-open it again by adding generic types to the array probably when Conditional Conformances are implemented in a future version of Swift.

Please let me know.

masters3d:
Let's do this implementation
https://rosettacode.org/wiki/Flatten_a_list#Swift

masters3d:
No so much the implemtatiom but the signature

`func flatten<T>(s: [Any]) -> [T] {`

AlwynC:
@masters3d Changed the signature for flattenArray method as you said in my latest commit.

exercises/flatten-array/FlattenArrayExample.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+func constructList(elements: Any?...) -> [Any?] {
```
regnerjr: I'm not convinced that this `constructList` function makes things more clear. 

`[1,2,3]` vs. `constructList(elements: 1, 2, 3)` ? What do you think?

Or maybe since it is used by the tests and not the implementation we should move it to be a helper function on the test target?

exercises/flatten-array/FlattenArrayExample.swift
```diff
@@ -0,0 +1,42 @@
...
+
+        for element in array {
+
+            if let j = element {
+
+                let anyObjectArray = j as? [Any?]
+
+                if let unwrappedArray = anyObjectArray {
+
+                    extractArrayElements(array: unwrappedArray)
```
regnerjr: I think we can remove the nested `extractArrayElements` function here and just recursively call the flatten array function. 

I see the reason you wanted/needed to do it this way is you are "closing over" the `flattenedArray` variable. ... both solutions are valuable.

exercises/flatten-array/FlattenArrayExample.swift
```diff
@@ -0,0 +1,42 @@
...
+
+func flattenArray<T>(_ list: [Any?]) -> [T] {
+
+    var flattenedArray = [T]()
+
+    func extractArrayElements(array: [Any?]) {
+
+        for element in array {
+
+            if let j = element {
```
regnerjr: What if `element` is nil... I don't see an `else` block...

Perhaps we could use `flatMap` on the array above to explicitly drop the `nil`s and only operate on items which exist? this would also reduce one level of nesting. 

`for element in array.flatmap{$0}` now every element we iterate over is known to be not `nil`.

exercises/flatten-array/FlattenArrayTest.swift
```diff
@@ -0,0 +1,37 @@
...
+        XCTAssertEqual([0, 2, 2, 3, 8, 100, -2], result)
+    }
+
+    func testFlattenForAllNullDeepNestedList() {
+
+        let nilValue: Any? = nil
+        let result: [Int] = flattenArray(constructList(elements: nilValue, constructList(elements: constructList(elements:constructList(elements: nilValue))), nilValue, nilValue, constructList(elements:constructList(elements:nilValue, nilValue), nilValue), nilValue))
+        XCTAssertEqual([], result)
+    }
+
```
regnerjr: It would be great to have at least one test case each which flattens `[String]` and `[Double]` just to enforce the `T` which we've written in our code.

AlwynC:
@regnerjr Please have look at my latest commit.

regnerjr:
This looks great @AlwynC . Nice work. 

We'll let @masters3d Give it the final 👍

masters3d:
Thank you!

PR #231, opened 2017-01-09
Fix SwiftLint warnings

robtimp:


robtimp:
@masters3d

regnerjr:
@robtimp Can we find an example of how the Swift team is formatting these? and choose to follow their style?

masters3d:
I am fine with disabling the dictionary part of the rule. I don't think there is a consensus on this on the swift library codebase.

robtimp:
@masters3d Ok, I'll open another PR to disable it.

Apple's Swift language book seems to use this format, but the language and standard library use either space on both sides or no spaces at all.

Issue #233, opened 2017-01-14
Swift Package manager


harquail:
Using [Swift PM](https://github.com/apple/swift-package-manager/tree/master/Documentation) for Swift would simplify the workflow for development (no need to create an Xcode project).  Running tests is also simple: https://github.com/apple/swift-evolution/blob/master/proposals/0019-package-manager-testing.md.

More info here: https://github.com/apple/swift-package-manager/blob/master/Documentation/Reference.md#test-suite-layouts

masters3d:
I agree. 
https://github.com/exercism/xswift/issues/142#issuecomment-266601381

PRs are welcomed.  ;)

lyuha:
@masters3d Must PRs include about all exercises with docs? Or split to PR including update one exercise?

masters3d:
You could make this one PR for all.  Docs get served separately. See sample in comment linked above.

masters3d:
https://github.com/masters3d/SwiftPackageManagerSamples/tree/master/Swift-PackageAccumulate

lyuha:
Ok. I update all exercises.

harquail:
thanks to you both for taking this on 😸

masters3d:
We need to update documentation before we can push into master.

harquail:
I can make a PR to update the documentation, unless @lyuha wants that task?

lyuha:
Good. I don't wants.

masters3d:
Closing this now. We just need to make all tests pass on Linux before merging into master

dgdosen:
Just started going through the swift exercises... and was going to suggest command line projects... but this is a much better idea.   Thanks for taking on this task!

Issue #142, opened 2016-04-20
Detach Swift from apple platform


masters3d:
Swift 2.2 is able to run on Ubuntu Linux. 
Microsoft just made their Ubuntu subsystem avalible to windows as Beta. 
https://insights.ubuntu.com/2016/04/14/howto-ubuntu-on-windows-2/

I believe the best time to transition would be when swift 3 becomes available (end of 2016)
We would need to update the instructions on how to run on Linux, Windows and macOS with our using xcode. 

Reference:
https://github.com/exercism/xswift/pull/141

kotp:
Also I believe Xcode should be less of a 'must' and more of a 'can' in that document.

masters3d:
Right now we are sort of dependent on Travis CI to add support for Swift Linux; once that happens we can move that way.

masters3d:
In an effort to start the detachment of swift from apple we will drop the requirement of Xcode. We will start by removing the hello world xcode project. 
https://github.com/exercism/xswift/pull/219

In a subsequent update we will add instructions on how to use the package manager to generate an xcode version for the people who want it. 

this will require us to change the internals of each 

```
Package
└── Sources
    ├── Foo.swift
    └── Tests
        └── Test.swift
```

from there we can generate an xcode project by running

``swift package generate-xcodeproj``

masters3d:
Working Sample:
https://github.com/masters3d/SwiftPackageManagerSamples/tree/master/Swift-PackageAccumulate

PR #236, opened 2017-01-16
first pass at updating docs for SPM

harquail:
Documentation update for #233 #142


harquail:
We could also update /docs/INSTALLATION.md to include the Linux installation steps, but I only use a mac so I wasn't sure of the steps.

masters3d:
lets keep the xcode documentation for now for people who want to use xcode manualy. 

lets add a new section on top for the package manager commans or links to official documentation. 
people can also easily create an xcode project from the command line using the package manager. 

what do you think?

harquail:
@masters3d  I can certainly add that in also so there are three possible paths (manually-created Xcode project, SPM-generated project, or just CLI + text editor), but it might be confusing for new users, who should probably just generate a project for each exercise

docs/TESTS.md
```diff
@@ -1,112 +1,65 @@
-## Xcode Tutorial
+## Testing Tutorial
 
-This guide explains how to create and configure a Swift-based Xcode project for Exercism.
+This guide explains how to run unit tests for Exercism Swift exercises, and optionally how to use Xcode for development.
 
-Apple does not support user-defined project templates in Xcode, and the default templates are more elaborate than those needed for Exercism. This guide bridges that gap.
+Exercism uses [Swift Package Manager](https://github.com/apple/swift-package-manager/tree/master/Documentation) to package files and tests for Swift.
```
harquail: add `swift test` and `swift package generate-xcodeproj` as a TLDR here for experienced users?

masters3d:
i was thinking we could include a little bash script that can generate and open an xcode project. 
we cant include the genereted project because the package manager creates hard links in the project so the project should not be checked into git.

docs/TESTS.md
masters3d: that would work.

harquail:
something like this? 
`swift package generate-xcodeproj |  grep -oE "[^ ]+$" | xargs open`

harquail:
although that may not be significantly easier than generating the project yourself :man_shrugging:

masters3d:
sounds good. documentation is tricky.

harquail:
added the TLDR — let me know if there are other changes or confusing steps 😃.  If we want to do the project generation bash script, that probably makes sense as a separate PR

docs/TESTS.md
bobdel: I think adding the Swift Package Manager is a big improvement. But I'm concerned the SPM edits made to Tests.md drops some important concepts that users who are coming in with a "beginners mindset" may find helpful. For example:

The context isn't clear that there are now two techniques for running the tests, Xcode and command line. Nor the consequences of each option. For example, the loss of inline compiler messages.

Running tests from the command line strips away error messages that might help new users, the command line output only displays messages from the console, not stderr. Depending on the situation, that missing info is valuable.

Bob

bobdel: I should have added, I'm coming at this purely from the Mac, and only recently began using SPM at all. I just worked though an example and think that tests.md is excellent, and the issues I noted above remain but are very simple (brief) to address. I also saw a few other sentences that might be out of place.

masters3d: I agree with you Bob. The issue here is that while xcode is an excellent IDE, it is only avalible on the mac. In order to be more inclusive we should strive give more options. When Swift was mac only it made sense to focus on xcode since this was the only way to work with swift. Learning xcode was part of learning Swift. This is still the case for the Objective-C track.  Do you have any specific changes you want to suggest? I know that you spent some time editing the current version. An alternative would be to pull the xcode detail instructions into a blog post that we could then link in here.

bobdel: My aim isn't to preserve the work I did before. The new version is much shorter and easier to follow. I think its fine to wipe it out. I also agree that Mac-only is less than ideal. 

Perhaps explaining that this track works under both Mac and Linux near the top.I don't think that is explicit enough yet. 

Then break out the Xcode specific stuff to the bottom on the document.

I'd be happy to suggest a revision, if that's welcome.

harquail: Let's make whatever changes we can to make this more approachable for both beginners and Linux users.  I think we clearly need to document both the Xcode and CLI use cases, but we can probably do more to explain why you would use one over the other.  @bobdel could you comment on specific lines of the changeset that are confusing?

masters3d: Absolutely. I appreciate your help on this.  We are working on branch so the changes are not going on master yet. I am going to accept this PR. Feel free to submit a new PR with your changes. Thank you!

dgdosen: Is there prior art for having alternative ways of "solving" exorcism problems?

masters3d: We are talking about the setup for each exercise, before we only gave instruction for xcode but now there are other ways like AppCode or https://nuclide.io/docs/languages/swift/

PR #244, opened 2017-01-19
Generalize docs

bobdel:
My initial take on revising the docs to reflect the use of the Swift Package Manager and alternatives to Xcode.


bobdel:
I think I did something wrong here. This should only contain 1 commit, with changes to three files; about.md, installation.md, and tests.md.

masters3d:
This looks great. I moved the target to our linux branch.

PR #250, opened 2017-01-24
Update tests.md

bobdel:
Edit the testing tutorial and reduced the size of images.


masters3d:
Thanks!

bobdel:
No problem!

Can I ask GitHub advice?

Before I started this latest update my fork on Github was equal to upstream. I made a new branch for the pull request and made all my changes there. 

Now my master in my fork is four commits behind. After studying Github for a bit, I've come to believe I cannot update my fork on Github directly. If I do, it'll add an extra merge commit that will put it ahead of upstream. Correct?

For my fork to match upstream on GH cleanly, I need to make sure my local working directory has no changes, then pull from upstream remote master and rebase with local master, then push to origin master also with rebase. That'll reset my fork to match upstream, without adding any extra merge commits. Does that sound right?

masters3d:
Oh yeah. What is happening is that if there are multiple commits in the PR, I usually just squish them together into one commit. This breaks the sync with your branch because now you have different commits than the exercism/master. Honestly I just check out a new branch from exercism/master for every PR. 

From my set up I do 
``
git pull exercism master
``

Sometimes is things go wrong I have to reset my branch. 

``
git reset exercism/master
``

There is also an option to hard reset but you should look that up because it is more destructive
.

bobdel:
Thanks! I'm making progress.

PR #251, opened 2017-01-25
Fix line endings in SETUP.md

DuBistKomisch:
Just started the course on exercism and there's a bunch of `^M`s in all the READMEs!


masters3d:
On the CLI?

What change did you do so we can let the client people know.

DuBistKomisch:
I use vim, so I can see the extra characters in the setup section of the readme of each problem.

If you open up this SETUP.md file it has dos line endings, so I just converted it to unix.

masters3d:
Thanks!

Issue #266, opened 2017-04-26
GradeSchool sorted school test


hobbitjoe:
When running the tests for GradeSchool I noticed that I had passed the test for checking the sortedRoster despite not actually doing any sorting.  Running my code through a repl I was able to see that the dictionary and contained arrays were not sorted.  Looking though the tests it looks like they may be incorrect.

Line 91:
`XCTAssertEqual(Array(result.keys).sorted(by: >), Array(expected.keys).sorted(by: >))`
The test is sorting the keys and then checking to see if they are equal.

Helper functions starting at line 6:
```// Workaround unit test liminitaiton on Optional Arrays/Sets
    private func sameCollection<C: Collection>(_ result: C?, _ expected: [String]?) -> Bool where C.Iterator.Element == String {
        guard let result = result, let expected = expected, similar(result, expected) else { return false }
        for (index, student) in result.enumerated() {
            guard index < expected.count && expected.contains(student) else { return false }
        }
        return true
    }
    private func similar<C: Collection>(_ result: C, _ expected: [String]) -> Bool {
        return result.count.toIntMax() == IntMax(expected.count)
    }

    func XCTAssertEqualCollection<C:Collection> (_ collectionS: C?, _ collectionA: [String]? ) where C.Iterator.Element == String {
        XCTAssert(sameCollection(collectionS, collectionA))
    }
```

These functions seem to just check to see if the number of items is correct and then if the student is contained within the expected list, not the order that the students are in.

masters3d:
Sets and dictionaries are non sorted data types. 
https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/CollectionTypes.html#//apple_ref/doc/uid/TP40014097-CH8-ID113

Keys can not be sorted, only arrays. Is this failing to check an arrays order?

hobbitjoe:
That makes sense that dictionaries are non sorted data types, but for the array order this line seems to only be looking to see if its contained:
```guard index < expected.count && expected.contains(student) else { return false }```

masters3d:
Yeah. This was added to be able to support `Set` data structure. I don't think is worth doing; we should just stick to `Array`. 
PR are welcome to remove this. 
Related: https://github.com/exercism/xswift/issues/259

PR #267, opened 2017-04-26
changed the test to use arrays for the sorted comparision.  Both the …

hobbitjoe:
changed the test to use arrays for the sorted comparision. Both the result and the expected array need to be equal in order to pass the test (if they are not equal the result array is not in the same order as the expected).  This is for issue #266

### DIFF
--- a/exercises/grade-school/Tests/GradeSchoolTests/GradeSchoolTests.swift
+++ b/exercises/grade-school/Tests/GradeSchoolTests/GradeSchoolTests.swift
@@ -3,68 +3,60 @@ import XCTest
 
 class GradeSchoolTests: XCTestCase {
 
-    // Workaround unit test liminitaiton on Optional Arrays/Sets
-    private func sameCollection<C: Collection>(_ result: C?, _ expected: [String]?) -> Bool where C.Iterator.Element == String {
-        guard let result = result, let expected = expected, similar(result, expected) else { return false }
-        for (index, student) in result.enumerated() {
-            guard index < expected.count && expected.contains(student) else { return false }
-        }
-        return true
-    }
-    private func similar<C: Collection>(_ result: C, _ expected: [String]) -> Bool {
-        return result.count.toIntMax() == IntMax(expected.count)
+    func sameArray(_ result: [String]?, _ expected: [String]? ) -> Bool {
+        guard let result = result, let expected = expected else { return false }
+        return result == expected
     }
 
-    func XCTAssertEqualCollection<C:Collection> (_ collectionS: C?, _ collectionA: [String]? ) where C.Iterator.Element == String {
-        XCTAssert(sameCollection(collectionS, collectionA))
+    func XCTAssertEqualArray (_ result: [String]?, _ expected: [String]? ) {
+        XCTAssertTrue(sameArray(expected, result))
     }
 
     func testAnEmptySchool() {
         let school   = GradeSchool()
         let result   = school.roster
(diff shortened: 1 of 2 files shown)


exercises/grade-school/Tests/GradeSchoolTests/GradeSchoolTests.swift
```diff
@@ -3,80 +3,72 @@ import XCTest
...
 
     func testAnEmptySchool() {
         let school   = GradeSchool()
         let result   = school.roster
         XCTAssertTrue(result.isEmpty)
-
     }
 
     func testAddStudent() {
-        var school = GradeSchool()
```
masters3d: We need these as var because people can use structs too.

exercises/grade-school/Tests/GradeSchoolTests/GradeSchoolTests.swift
```diff
@@ -3,80 +3,72 @@ import XCTest
...
-    private func similar<C: Collection>(_ result: C, _ expected: [String]) -> Bool {
-        return result.count.toIntMax() == IntMax(expected.count)
+    func sameArray(_ result: [String]?, _ expected: [String]? ) -> Bool {
+        guard let result = result, let expected = expected else { return false }
+        return result == expected
     }
 
-    func XCTAssertEqualCollection<C:Collection> (_ collectionS: C?, _ collectionA: [String]? ) where C.Iterator.Element == String {
-        XCTAssert(sameCollection(collectionS, collectionA))
+    func XCTAssertEqualArray (_ result: [String]?, _ expected: [String]? ) {
```
masters3d: You can actually just use an XCTAssertEqual here, we only had to provide our own when we made that generic funcion

hobbitjoe: I was doing that before but I needed to unwrap the value in cases like:
```XCTAssertEqual(result[3], expected[3])```

Because of that I used the helper functions to safely unwrap and do the assertion without having guards in every function that accesses the students in the grade.

masters3d: sounds good.

masters3d:
Do you mind changing the Example so that the test pass?
https://github.com/hobbitjoe/xswift/blob/master/exercises/grade-school/Sources/GradeSchoolExample.swift

Right now we are using the Set but they need to be Array

masters3d:
Do you mind addressing the Lint errors

⚠️	/exercises/grade-school/Sources/GradeSchoolExample.swift#L2 - Shorthand syntactic sugar should be used, i.e. [Int] instead of Array
⚠️	/exercises/grade-school/Sources/GradeSchoolExample.swift#L14 - Shorthand syntactic sugar should be used, i.e. [Int] instead of Array
⚠️	/exercises/grade-school/Sources/GradeSchoolExample.swift#L15 - Shorthand syntactic sugar should be used, i.e. [Int] instead of Array
⚠️	/exercises/grade-school/Sources/GradeSchoolExample.swift#L18 - Shorthand syntactic sugar should be used, i.e. [Int] instead of Array
⚠️	/exercises/grade-school/Sources/GradeSchoolExample.swift#L19 - Shorthand syntactic sugar should be used, i.e. [Int] instead of Array
⚠️	/exercises/grade-school/Sources/GradeSchoolExample.swift#L25 - Files should have a single trailing newline.

masters3d:
Thanks!

PR #279, opened 2017-07-06
Rna transcription fix

samkrishna:
The failure cases defined in `canonical-data.json` were not included in the original test code.

As a result, the one complex error case failed ("ACGTXXXCTTAA") when the original code returned a truncated version of the transcription, ignoring the "XXX" middle.

I've added the tests and fixed the code. It now returns `nil` as per the spec.

### DIFF
--- a/exercises/rna-transcription/Sources/RnaTranscriptionExample.swift
+++ b/exercises/rna-transcription/Sources/RnaTranscriptionExample.swift
@@ -1,6 +1,11 @@
-struct Nucleotide {
+enum TranscriptionError: Error {
+    case InvalidNucleotide
+}
 
-    var complementOfDNA: String? { return transcribe(dnaToRna) }
+struct Nucleotide {
+    func complementOfDNA() throws -> String {
+        return try transcribe(dnaToRna)
+    }
 
     private let value: String
 
@@ -10,11 +15,11 @@ struct Nucleotide {
 
     private let dnaToRna: [Character:String] = [ "G": "C", "C": "G", "T": "A", "A": "U" ]
 
-    private func transcribe(_ dict: [Character : String]) -> String? {
+    private func transcribe(_ dict: [Character : String]) throws -> String {
         var tempText = ""
         for each in self.value.characters {
             if (dict[each] == nil) {
-                return nil
+                throw TranscriptionError.InvalidNucleotide
             }
             tempText += dict[each] ?? ""
         }
--- a/exercises/rna-transcription/Sources/RnaTranscriptionExample.swift
+++ b/exercises/rna-transcription/Sources/RnaTranscriptionExample.swift
@@ -1,6 +1,6 @@
 struct Nucleotide {
 
-    var complementOfDNA: String { return transcribe(dnaToRna) }
+    var complementOfDNA: String? { return transcribe(dnaToRna) }
 
     private let value: String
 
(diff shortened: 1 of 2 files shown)


masters3d:
I don't think we need a nullable String here. The way that I've seen null handled for Array is an empty array []; the same goes for String ( which in Swift also a collection ) is an empty String "".  I've be fine with this being a throwing function though.

samkrishna:
OK that's a good point -- I remember reading that somewhere in the Swift 4 Language Book. I'm still getting used to Swift being almost as throw-happy as Java, since I'm coming from pure Objective-C development.

I'll make the mods to throw and push.

samkrishna:
OK I've pushed my changes. 

@masters3d Could you take a look at the test function `testInvalidRnaComplementOfUracil()` ? I've made an attempt to call `XCTAssertThrowsError()` with a follow-up to `XCTAssertEquals()` to attempt to verify that the correct error was being returned, but couldn't figure out how to get the code to work.

Thanks!

exercises/rna-transcription/Sources/RnaTranscriptionExample.swift
```diff
@@ -10,9 +15,12 @@ struct Nucleotide {
 
     private let dnaToRna: [Character:String] = [ "G": "C", "C": "G", "T": "A", "A": "U" ]
 
-    private func transcribe(_ dict: [Character : String]) -> String {
+    private func transcribe(_ dict: [Character : String]) throws -> String {
         var tempText = ""
         for each in self.value.characters {
+            if (dict[each] == nil) {
```
masters3d: The linter is complaining here. Run the autocorrect to see if some of the others go away.

exercises/rna-transcription/Tests/RnaTranscriptionTests/RnaTranscriptionTests.swift
```diff
@@ -3,32 +3,51 @@ import XCTest
...
     func testRnaComplement() {
-        XCTAssertEqual("UGCACCAGAAUU", Nucleotide("ACGTGGTCTTAA").complementOfDNA)
+        XCTAssertEqual("UGCACCAGAAUU", try Nucleotide("ACGTGGTCTTAA").complementOfDNA())
     }
-
+    
+    func testInvalidRnaComplementOfUracil() {
+        XCTAssertThrowsError(try Nucleotide("U").complementOfDNA())
+        
+//        XCTAssertThrowsError(try Nucleotide("U").complementOfDNA() { (error) -> Void in
```
masters3d: Let's remove this comment.

masters3d:
Swift errrors are untyped.

exercises/rna-transcription/Sources/RnaTranscriptionExample.swift
samkrishna:
```
$ SwiftLint autocorrect
Correcting Swift files in current working directory
Correcting 'RnaTranscription.swift' (1/2)
Correcting 'RnaTranscriptionExample.swift' (2/2)
Done correcting 2 files!
```

samkrishna:
OK -- so you want me to make the errors untyped? Not a problem.

masters3d:
Since there is only one case, I think untyped here is good.

samkrishna:
OK it looks like it's all done. LMK if there's something else you need from me here on this exercise.

PR #281, opened 2017-07-06
Fixed a few more linting warnings

samkrishna:
Added a few more fixings for various linting warnings that missed the 1st pass

### DIFF
--- a/exercises/custom-set/Sources/CustomSetExample.swift
+++ b/exercises/custom-set/Sources/CustomSetExample.swift
@@ -2,7 +2,7 @@ func == <T> (lh: CustomSet<T>, rh: CustomSet<T>) -> Bool {
     return lh.contents.keys.sorted { $0.hashValue < $1.hashValue } == rh.contents.keys.sorted { $0.hashValue < $1.hashValue }
 
 }
-struct CustomSet<T:Hashable>:Equatable {
+struct CustomSet<T:Hashable>: Equatable {
 
     typealias Element = T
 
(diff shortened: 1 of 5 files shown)


samkrishna:
For @masters3d:

Also: what about migrating the codebase to Swift 4? I see that Xcode9b2 is offering to migrate the Swift 3 code. Any thoughts?

samkrishna:
@masters3d 

Also: the linting warning that triggers in the PokerExample.swift code is here:

https://www.natashatherobot.com/swift-unused-optional-value/

Unfortunately, I don't see how to get around this in the case statements w/o the build breaking. Any suggestions?

masters3d:
Are you referring to? `Arguments can be omitted when matching enums with associated types if they are not used.`

samkrishna:
@masters3d Yes.

masters3d:
This seems to work:

```swift
    func order() -> Int {
        switch self {
        case .highCard: return 1
        case .onePair: return 2
        case .twoPair: return 3
        case .threeOfAKind: return 4
        case .straight: return 5
        case .flush: return 6
        case .fullHouse: return 7
        case .fourOfAKind: return 8
        case .straightFlush: return 9
        }
    }
```

masters3d:
I merged [this](https://github.com/exercism/swift/pull/278) to address the swift4 and swift3 compatability

samkrishna:
@masters3d Your switch-case code definitely works in Swift 3.x on Xcode 8.3.3.

However, it breaks my build on Xcode 9b2.

masters3d:
I am using the the 4.0 swift toolchain that is a little newer than 9b2 https://swift.org/download/#snapshots

samkrishna:
OK I'm going to download that version. Hopefully 9b3 will have what you're seeing.

masters3d:
we need to support the current version though which is still 3.1

samkrishna:
OK I've installed the Swift 4.0 Development Toolchain and switched to it. The Swift 3.1.1 code works under it. It still gives the annoying "Switch must be exhaustive" warning but it does build with that code and passes all tests.

PR #295, opened 2017-09-06
Add test for other allergies

davidahouse:
This PR updates the [Allergies](http://exercism.io/exercises/swift/allergies/readme) exercise and adds a new unit test. This test covers one of the stated criteria: 

`Note: a given score may include allergens not listed above (i.e. allergens that score 256, 512, 1024, etc.). Your program should ignore those components of the score. For example, if the allergy score is 257, your program should only report the eggs (1) allergy.`

The test just covers the case of a score of 257.


masters3d:
Thanks!

PR #298, opened 2017-09-30
Added Test for sumOfMultiples inc ase limit is 0 and there is a 0 amo…

cham-s:
Added a new edge case when the limit is zero and there is zero in a set of multiples.


masters3d:
Do you mind editing the sample code to this passes CI?

cham-s:
Edited Example code with a guard statement.

masters3d:
Thanks!

PR #299, opened 2017-10-10
Rewrote tests to follow canonical data; reworked example to pass tests (fixes #232)

### LINKED ISSUE #232: Changes to Luhn tests
We recently rewrote the test suite for the Luhn exercise. Since this track implements Luhn, please [take a look at the new canonical_data.json file](https://github.com/exercism/x-common/blob/master/exercises/luhn/canonical-data.json) and see if your track should update its implementation or tests.

Details on the test changes and the motivation can be found in [the PR](https://github.com/exercism/x-common/pull/474).

Tracking of these changes is being done at [the tracking issue](https://github.com/exercism/x-common/issues/491).

jarhill0:
The canonical data uses strings as input rather than ints, which required a partial rework of the example. Also, the example uses `>` where it should use `>=`, leading to a bug whenever a number contained the digit 5. I also removed an unused function from the example.

### DIFF
--- a/exercises/luhn/Sources/LuhnExample.swift
+++ b/exercises/luhn/Sources/LuhnExample.swift
@@ -3,32 +3,25 @@ import Foundation
 struct Luhn {
 
     var number: Int64 = 0
-    var addends: [Int] { return addendsFunc(number) }
+    var numberString: String = ""
+    var addends: [Int] { return addendsFunc(numberString) }
     var checksum: Int { return addends.reduce(0, +) }
-    var isValid: Bool { return checksum % 10 == 0 }
+    var isValid: Bool = false
 
-    init(_ num: Int64) {
-        self.number = num
-    }
-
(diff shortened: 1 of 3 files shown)


masters3d:
Looks great. Swift 4 is so new so we are still supporting swift 3.1. Do you mind taking a look at the CI?

jarhill0:
@masters3d How would you recommend finding the length of a string in swift 3.1 so that the test will pass? The internet has several conflicting answers about what is best.

masters3d:
mystring.utf16.count should do the trick

https://github.com/exercism/swift/blob/c9c9014d36a45243b608b8324c9dd84060dc4d92/exercises/minesweeper/Sources/MinesweeperExample.swift#L123

jarhill0:
Updated.

masters3d:
Thanks!

PR #306, opened 2017-11-22
Add ISBN Verifier exercise

robtimp:
* Add ISBN Verifier exercise in Swift

### DIFF
--- a/config.json
+++ b/config.json
@@ -360,6 +360,19 @@
     },
     {
       "core": false,
+      "difficulty": 3,
+      "slug": "isbn-verifier",
+      "topics": [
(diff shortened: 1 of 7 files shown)


robtimp:
@masters3d

robtimp:
@masters3d Any idea why the Linux Swift 4 build fails?

masters3d:
I guess we need to update the docker image:
https://hub.docker.com/r/norionomura/swift/tags/

masters3d:
We should probably use official images
https://hub.docker.com/r/library/swift/tags/
4.0.2 looks good.

PR #308, opened 2017-11-28
Fix swift 4 docker

masters3d:
Fix #306 
Fix #307


robtimp:
@masters3d Thank you!

PR #312, opened 2017-12-04
Add Protein Translation exercise

robtimp:
* Add Protein Translation in Swift


masters3d:
I’d be fine with removing swift 3 comparability. 
We should also remove the Mac build from Travis. It takes too long.

Issue #309, opened 2017-11-28
Travis CI way too slow on mac. 


masters3d:
Look into deprecating Travis CI for mac and only use for Linux. 

![image](https://user-images.githubusercontent.com/6539412/33304051-632c66dc-d3bb-11e7-9aee-eb8966766159.png)

PR #317, opened 2018-01-16
moved danger and macOS to circle

masters3d:
- Steps to remove travisCI mac

PR #318, opened 2018-01-16
try to fix circle format

masters3d:
https://github.com/exercism/swift/pull/317


masters3d:
closes https://github.com/exercism/swift/issues/309

PR #320, opened 2018-01-16
 Don't fail build on non zero exit codes

masters3d:
Don't fail build on danger errors


masters3d:
Merge. Travis is not working well.
```
Investigating - We are continuing to work on resolving this issue, and are investigating other potential causes for the delay. 
Jan 16, 21:17 UTC
```
https://www.traviscistatus.com

PR #325, opened 2018-02-09
Add Scale Generator exercise

robtimp:
* Add Scale Generator exercise in Swift


robtimp:
@masters3d I'm tempted to leave this as is if we're dropping Swift 3 soon.

masters3d:
Lets do it. We will need to update the package format at some point but we can start tracking 4.1x changes now. https://github.com/exercism/swift/pull/326

masters3d:
Merge when you are ready

PR #330, opened 2018-03-12
Add the missing .gitignore in exercises

bcylin:
Some of the exercises are missing `.gitignore` to exclude the swift package manager generated xcodeproj. This PR adds the missing files and updates Dangerfile to check future changes if needed.


masters3d:
Thanks! I apologize about the delay.

PR #337, opened 2018-04-17
Add Proverb Exercise

AlwynC:
Proverb exercise implemented in Swift,
Swiftlint autocorrect applied to the project.


exercises/proverb/Sources/ProverbExample.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+struct ProverbExample {
+
+    func recite(pieces: [String]) -> String {
+
+    var saying = String()
+
+    if !pieces.isEmpty {
```
robtimp: Changing this to `if let first = pieces.first {` will also ensure that pieces is not empty, while avoiding the force unwrap below.

exercises/proverb/Sources/ProverbExample.swift
```diff
@@ -0,0 +1,28 @@
...
+
+        for (position, piece) in pieces.enumerated() {
+
+            if position != pieces.index(pieces.endIndex, offsetBy: -1) {
```
robtimp: Could be simplified to `if position != pieces.count - 2 {`

exercises/say/.build/regenerate-token
```diff
@@ -0,0 +1 @@
+0
```
masters3d: Can you remove all *-token files from this PR?

exercises/proverb/Tests/ProverbTests/ProverbTests.swift
```diff
@@ -0,0 +1,73 @@
+import XCTest
+@testable import Proverb
+
+class ProverbTests: XCTestCase {
+
+    var proverb =  Proverb()
```
masters3d: looks good.

exercises/proverb/Tests/ProverbTests/ProverbTests.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    override func setUp() {
+        super.setUp()
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDown() {
```
masters3d: you dont need these `tearDown` or `setUp`

exercises/proverb/Tests/ProverbTests/ProverbTests.swift
```diff
@@ -0,0 +1,73 @@
...
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+     super.tearDown()
+    }
+
+    func testZeroPiecesOfProverb() {
+
+        let expectedResult = String()
```
masters3d: initialize to an empty string instead. `let expectedResult = ""`

exercises/proverb/Sources/ProverbExample.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+struct Proverb {
+
+    func recite(_ pieces: [String]) -> String {
```
masters3d: This should probably be an extension on `Array where Element:String`

exercises/proverb/Sources/ProverbExample.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    var saying = String()
+
+    if let firstPiece = pieces.first {
```
masters3d: You can check if the array is empty instead of checking is it has a first element

robtimp: This was my suggestion. It avoids the need to force unwrap the first element when it's used below

exercises/proverb/Sources/ProverbExample.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+struct Proverb {
+
+    func recite(_ pieces: [String]) -> String {
+
+    var saying = String()
```
robtimp: I think this would be more idiomatic as `""` (vs. `String()`) as well

exercises/proverb/Sources/ProverbExample.swift
masters3d: sounds good! Usually I do something like `guard !array.isEmpty() else {...}` but it really comes down to style

exercises/proverb/Sources/ProverbExample.swift
AlwynC: I don't feel this is necessary...

AlwynC: But it can be done if needed...

AlwynC: Ok I did it.

robtimp:
@AlwynC Looks good! Thanks

PR #338, opened 2018-04-19
Compare floating point with tolerance in space age test.

kellegous:
The test for space age seems to do a direct comparison of floating point numbers. I don't know swift very well at all, but with IEEE 754 floating point numbers one typically needs to compare them with a tolerance of error to account for small rounding errors introduced through arithmetic operations. This just adds such a tolerance to each of the `XCTAssertEqual` calls in the SpaceAge tests.


robtimp:
Thanks @kellegous! This is a good improvement to these tests

PR #341, opened 2018-05-16
robot-name: test for upper-case letters

paiv:
Name format is "two uppercase letters followed by three digits".

In this patch, regex `[A-Z]{2}` tests for two upper-case ASCII letters.

Original regex `\w{2}` expands to `[a-zA-Z0-9_]{2}` which is somewhat relaxed.

Alternative regex `\\p{Lu}{2}` could accept any Unicode upper-case letter.


masters3d:
thanks!

PR #342, opened 2018-05-21
scale-generator: fix enigmatic scale pattern

paiv:
This fixes the scale to 12 half steps, as in [canonical-data](https://github.com/exercism/problem-specifications/blob/c85397312197906905c295ec7d026e53159d2fd4/exercises/scale-generator/canonical-data.json#L169-L177)


paiv:
Also see how python added a test for broken interval:

https://github.com/exercism/python/blob/6d13ec69a978e4920bc326b666814d2770aa5c74/exercises/scale-generator/scale_generator_test.py#L82-L85

robtimp:
@paiv Good catch!

PR #351, opened 2018-07-18
added blurb

masters3d:
added summary of language

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "language": "Swift",
   "active": true,
-  "blurb": "",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -1,7 +1,7 @@
 {
   "language": "Swift",
   "active": true,
-  "blurb": "",
+  "blurb": "Swift is a modern open source programming language that runs on macOS, Linux and on Windows using a linux subsystem.",
   "exercises": [
```
kytrinyx: I'd love to have something a bit more approachable here than where the language runs.

E.g. the blurb for R specifically calls out the powerful data manipulation and data science applications.

masters3d:
@kytrinyx  what do you think?

kytrinyx:
This looks great!

Issue #356, opened 2018-08-07
Rna Transcript Tests that test throwing functions with error messages are confusing


Zialus:
When solving the exercise `Rna Transcript` I found the test suite to be a bit strange.

Specifically, I found confusing the message parameter used in the [XCTAssertThrowsError function](https://github.com/exercism/swift/blob/master/exercises/rna-transcription/Tests/RnaTranscriptionTests/RnaTranscriptionTests.swift#L25-L50)

`XCTAssertThrowsError(try Nucleotide("XXX").complementOfDNA(), "This didn't work")`

`_ expression: @autoclosure () throws -> T, _ message: @autoclosure () -> String = default`

[Here's the documentation for XCTAssertThrowsError](https://developer.apple.com/documentation/xctest/1500795-xctassertthrowserror)

I asked the mentor that reviewed the exercise about it, but he didn't comment on it.

Wouldn't it be better to change the tests to something like what is done in the [Grains](https://github.com/exercism/swift/blob/master/exercises/grains/Tests/GrainsTests/GrainsTests.swift#L6-L34) exercise?

masters3d:
Good point. Pull request welcome. Thanks!

Zialus:
Sure, give me some minutes.

PR #357, opened 2018-08-07
Improvements to Rna Transcript Tests

Zialus:
Solution to the problem described here https://github.com/exercism/swift/issues/356


masters3d:
Thank you!

PR #360, opened 2018-08-30
difference-of-squares: renamed squareOfSums to squareOfSum

Scientifica96:
In exercise "difference-of-squares", renamed `squareOfSums` to `squareOfSum`.


masters3d:
Thanks!

Scientifica96:
It's a pleasure :D

PR #362, opened 2018-10-01
Fix warnings related to old String's character view property

bhargavg:
In latest version of Swift, `.characters` property is deprecated. This patch helps remove those deprecation warnings.

> Please note that code can be refactored in few places using the latest Swift's String APIs. This can be done in a separate PR.


masters3d:
Thanks!

PR #366, opened 2018-10-04
Fix swiftlint not running on circle CI

bhargavg:
The existing commands doesn't try to install swiftlint if it's not already installed, causing linting to fail.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@ import PackageDescription
 import Foundation
 
 extension String {
-	var PascalCased: String {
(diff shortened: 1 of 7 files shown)


bhargavg:
`swiftlint` is enabled now.

> Please note that, I've disabled `cyclomatic_complexity` only for [this method](https://github.com/exercism/swift/blob/22a1e1b5c76c312999246c13f410464665574654/exercises/poker/Sources/Poker/PokerExample.swift#L189)

PR #369, opened 2018-10-24
Update example solution for hamming distance demonstrating filtering.

namratabandekar:
The hamming exercise is tagged with filtering. The example solution should use swift features and filtering to demonstrate this.

### DIFF
--- a/exercises/hamming/Sources/Hamming/HammingExample.swift
+++ b/exercises/hamming/Sources/Hamming/HammingExample.swift
@@ -1,25 +1,7 @@
-struct Hamming {
-    static func compute(_ input: String, against: String) -> Int? {
-        var differences = 0
-
-        let char1 = convertStringToArray(input)
-        let char2 = convertStringToArray(against)
-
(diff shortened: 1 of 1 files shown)


exercises/hamming/Sources/Hamming/HammingExample.swift
```diff
@@ -1,25 +1,7 @@
...
-
-private func convertStringToArray(_ input: String) -> [Character] {
-    var characterArray: [Character] = []
-
-    for character in input {
-        characterArray.append(character)
-    }
-
-    return characterArray
+func compute(_ dnaSequence: String, against: String) -> Int? {
```
masters3d: How interesting. Since the module is called Hamming, the tests did not break when called like this `Hamming .compute(...)`. I sort of like it.

masters3d:
Can yo address the linting message?

namratabandekar:
Thanks masters3d. Fixed the error.

masters3d:
Thank you

PR #377, opened 2019-03-14
Instructions for Xcode project creation added to readme

ThomasHaz:


ThomasHaz:
The only files changed were the README.md in the exercises. Same change applied to all of them to let the user know how to generate an Xcode project from source. (The existing README.md's don't point to the correct page for setting up Xcode [neither directly nor indirectly].)

complex-numbers had the whole `Setup` section added as it was the only one missing it altogether.

masters3d:
Looks good. What do think of the below project? https://swiftxcode.github.io/#what-it-looks-like-

PR #385, opened 2019-09-12
Remove test cases that break tie by suit

cdmcmahon:
**Proposal:** Remove test cases that enforce a winner among identically hands in different suits.

**Reasoning:** According to the rules of poker, there is no ranking of suits. Two equal hands with differing suits are considered to tie.

[Source 1 - World Series of Poker:](https://www.wsop.com/poker-hands/)
> The suits are all of equal value - no suit is higher than any other suit.

[Source 2 - Wikipedia (referenced in the exercise description):](https://en.wikipedia.org/wiki/List_of_poker_hands)
> Suits are not ranked, so hands that differ by suit alone are of equal rank. 

**Alternatives:** We could add tests to test for ties, perhaps by having the bestHands method return nil in such cases.

### DIFF
--- a/exercises/poker/Tests/PokerTests/PokerTests.swift
+++ b/exercises/poker/Tests/PokerTests/PokerTests.swift
@@ -216,7 +216,7 @@ class PokerTests: XCTestCase {
                     "Q♢ K♢ J♢ 10♢ A♢"
                 ],
                 best:  "Q♢ K♢ J♢ 10♢ A♢"
-            ),
+            )
         ]
 
         invalidTestCases  =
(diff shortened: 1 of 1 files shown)


masters3d:
I would prefer it if we replaced it with high card by suit https://en.m.wikipedia.org/wiki/High_card_by_suit

cdmcmahon:
[The linked Wikipedia page even includes:](https://en.m.wikipedia.org/wiki/High_card_by_suit)

> Most poker games do not rank suits; the ace of clubs is just as good as the ace of spades.

It feels like this is just trying to include another logical twist to the solution, even though it's not accurate to the problem domain.

PR #389, opened 2020-05-10
Added two tests and removed force-unwrapping in test cases

wneumann:
Added two tests that I have had students trip up on. Single nucleotide strands, and strands that contain both valid and invalid nucleotides (current validation test was only with all invalid characters).

More importantly, I modified the tests to remove force unwrapping and to use XCTUnwrap instead. We would encourage our students to not use force unwrapping in their code, and we shouldn't use it in ours.

### DIFF
--- a/exercises/nucleotide-count/.gitignore
+++ b/exercises/nucleotide-count/.gitignore
@@ -2,3 +2,4 @@
 /.build
 /Packages
 /*.xcodeproj
+.swiftpm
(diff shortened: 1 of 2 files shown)


wneumann:
Didn't realize that XCTUnwrap wasn't available on the system -- apparently it's Xcode only at the moment. Will rewrite with what is available.

PR #391, opened 2020-05-26
Add wneumann to swift maintainers list

wneumann:
Adding myself to the swift maintainers.json per @ErikSchierboom. Also added .swiftpm directory to the swift .gitignore.


iHiD:
Merging this so you get in the right GH teams sooner rather than later.

PR #399, opened 2020-11-05
Remove unintended whitespace in SaddlePointsTests.swift

thomchop:
The extra space was easy enough to parse out, but as a sole anomaly not alluded to in the README, it appeared unintentional. When it produced an error for me, the source wasn't readily apparent to me, so it's likely frustrated other learners too.


kytrinyx:
Thanks! The instructions were synced with the upstream problem-specifications in #586. Closing.

PR #404, opened 2020-11-06
Revise custom-set tests to comply with problem-spec

### LINKED ISSUE #146: Updated tests for the Custom Set problem
In order to reduce the amount of code required to pass incremental tests (assuming that users pass tests starting from the top), the order of the tests was modified slightly.

Since this track implements Custom Set, please take a look at the new custom-set.json file and see if your track should update its tests.

If you do need to update your tests, please refer to this issue in your PR. That helps us see which tracks still need to update their tests.

If your track is already up to date, go ahead and close this issue.

More details on this change are available in https://github.com/exercism/x-common/pull/257.

thomchop:
Intended to resolve #146 
To honor the problem spec, it seemed clear that a complete overhaul was needed, not just a reordering.

### DIFF
--- a/exercises/custom-set/Sources/CustomSet/CustomSetExample.swift
+++ b/exercises/custom-set/Sources/CustomSet/CustomSetExample.swift
@@ -14,6 +14,7 @@ struct CustomSet<T: Hashable>: Equatable {
 
     fileprivate var contents = [Element: Bool]()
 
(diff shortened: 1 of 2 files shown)


meatball133:
This pr is incompatible with how the current architecture currently looks like and due to its age I don't think the changes are relevant anymore.

PR #459, opened 2021-06-23
Add tags

angelikatyborska:
I want to help out with Exercism v3, so even thought I don't know Swift, I tried to pick out its tags. I based my selection on the wikipedia article (https://en.wikipedia.org/wiki/Swift_(programming_language)).

All available tags: https://github.com/exercism/docs/blob/main/building/tracks/config-json.md#tags

Closes https://github.com/exercism/swift/issues/427

### DIFF
--- a/config.json
+++ b/config.json
@@ -1490,5 +1490,23 @@
     }
   ],
   "key_features": [],
-  "tags": []
+  "tags": [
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
LGTM

PR #481, opened 2021-09-04
Fix expected value in totalTimeInMinutes

### LINKED ISSUE #480: lasagna test does not match spec
From the spec, `3 * 2 + 20 = 26`
https://github.com/exercism/swift/blob/e476a4a070ddb187d0203d2ed1b218d4a0f103e4/exercises/concept/lasagna/.docs/instructions.md#L36-L41

In this test the expected value should be `6 * 2 + 13 = 25`
https://github.com/exercism/swift/blob/e476a4a070ddb187d0203d2ed1b218d4a0f103e4/exercises/concept/lasagna/Tests/LasagnaTests/LasagnaTests.swift#L21-L24

paiv:
Fixes #480


ErikSchierboom (commented):
Great work! Could you also update the exemplar implementation: https://github.com/exercism/swift/blob/main/exercises/concept/lasagna/.meta/Sources/Lasagna/LasagnaExemplar.swift#L12

ErikSchierboom (approved):

PR #484, opened 2021-09-24
Fix prerequisites of concept exercises

sanderploegsma:
This fixes some prerequisites of concept exercises that point to concepts that do not exist (yet). Specifically, the following non-existing concepts were prerequisites for other exercises:

- `strings` (should probably be `characters` instead)
- `conditionals` (covered by `conditionals-if`, `conditionals-guard` and `conditionals-switch`)

Because of these incorrect prerequisites some exercises remain locked without a way to unlock them.

Note: An alternative fix for `conditionals` could be to add the `conditionals` concept to the appropriate exercises instead of changing the prerequisites of the others.

### DIFF
--- a/config.json
+++ b/config.json
@@ -143,7 +143,7 @@
         "name": "Santa's Swifty Helper",
         "uuid": "f371c072-082b-496b-aa58-45b6cba203f6",
         "concepts": ["tuples"],
-        "prerequisites": ["arrays", "strings"],
+        "prerequisites": ["arrays", "characters"],
         "status": "active"
       },
       {
--- a/config.json
+++ b/config.json
@@ -89,7 +89,7 @@
           "inout-parameters",
           "nested-functions"
         ],
-        "prerequisites": ["arrays", "conditionals", "tuples"],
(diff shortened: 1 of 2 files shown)


sanderploegsma:
Also, using separate concepts for if/guard/switch conditionals seems like overkill to me. In my opinion one single "conditionals" concept, where some exercises dive deeper into either `if`, `guard` or `switch` conditionals would be a cleaner way to do it, from a learner's perspective.

ErikSchierboom (approved):
> Also, using separate concepts for if/guard/switch conditionals seems like overkill to me. In my opinion one single "conditionals" concept, where some exercises dive deeper into either if, guard or switch conditionals would be a cleaner way to do it, from a learner's perspective.

Yep, this is totally fine. Feel free to change this in a follow-up PR.

sanderploegsma:
Any chance this can be merged? That allows me to continue on my Swift track 🙃

ErikSchierboom (approved):
Sorry, forgot about this

sanderploegsma:
No problem, thanks for merging! 👍

PR #487, opened 2021-09-27
Run concept exercises in CI

### LINKED ISSUE #469: Test Concept Exercises in CI
The CI workflow currently only tests practice exercises, but should also test concept exercises.

sanderploegsma:
This fixes #469 by adding the concept exercises to the root `Package.swift`. 

To get this working one change to the exercises was required: the exemplar implementations were all in `.meta/Exemplar.swift` files, which caused the exercise's tests to fail on the imports. I therefore moved all exemplar implementations to `.meta/Sources/[ExerciseName]/[ExerciseName]Exemplar.swift` and updated the `config.json` files accordingly.

Note that I removed the creation of `LinuxMain.swift` in the SPM manifest since the CI job no longer runs on Linux, so this is not necessary.


config.json
```diff
@@ -22,7 +22,7 @@
     "solution": ["Sources/%{pascal_slug}/%{pascal_slug}.swift"],
     "test": ["Tests/%{pascal_slug}Tests/%{pascal_slug}Tests.swift"],
     "example": ["Sources/%{pascal_slug}/%{pascal_slug}Example.swift"],
```
ErikSchierboom, line 24: I think this also needed to be updated to `[".meta/Sources/%{pascal_slug}.swift"]` Probably forgotten in a previous PR.

config.json
```diff
@@ -22,7 +22,7 @@
...
-    "exemplar": [".meta/Exemplar.swift"]
+    "exemplar": [".meta/ExemplarSources/%{pascal_slug}.swift"]
```
ErikSchierboom, line 25: Was there any particular reason not yo use `.meta/Sources` like the Practice Exercises?

ErikSchierboom (commented):
Fantastic! I've left two minor comments.

config.json
sanderploegsma: Not really, I mainly wanted to make it explicit that the folder contained an exemplar implementation, since the file name no longer contained the `Exemplar` term. 

I vaguely recall that I read in the documentation that a concept exercise could contain more than one example implementation, which would make the directory naming more important, but reading through it again I can’t make out where I got that idea.

I think using `.meta/Sources` like the practice exercises makes the most sense, so I’ll push an update to fix it.

ErikSchierboom: Great! There can only be one exemplar solution (it can consist of multiple files though).

sanderploegsma: I just noticed that the practice exercises actually use `.meta/Sources/%{pascal_slug}/%{pascal_slug}Example.swift`, so I think the concept exercises might use that same pattern as well, so `.meta/Sources/%{pascal_slug}/%{pascal_slug}Exemplar.swift`.

config.json
sanderploegsma: Looks like the example files are in `[".meta/Sources/%{pascal_slug}/%{pascal_slug}Example.swift"]`, so I updated the config to use that as default.

ErikSchierboom (approved):
Sweet. I also really like how the `xswift-test-spm` file is no longer needed.

ErikSchierboom:
@sanderploegsma We have a passing CI system again!

PR #490, opened 2021-10-04
Fix description of totalTimeInMinutes method for lasagna task

2Grey:
Changed description of lasagna task for test compliance.

In the task, we expect that in the `totalTimeInMinutes` method we will use the code of the previously implemented methods` remainingMinutesInOven` and` preparationTimeInMinutes`


2Grey:
@paiv, what do you think about this fix.

I think that is not "lasagna test does not match spec", most likely the task description is incorrect.

PR #495, opened 2021-10-06
Windowing System : changed instructions' Size width and height to make them fit tests.

adauguet:
Point 1 states to implement `resize(newHeight:newWidth:)` but tests are using `resize(newWidth:newHeight:)`.
Per convention `width` should go before `height`.


SleeplessByte (approved):
Thanks! 

Makes sense.

PR #500, opened 2021-10-18
lasagna-master: Update tests of final task to match instructions

sanderploegsma:
From the instructions:

> You've heard that you should serve white wine if there is more mozzarella, ricotta, and béchamel in the lasagna than there is meat and sauce, and red wine otherwise.

Let's call the number of mozzarella/ricotta/béchamel layers `A` and the number of meat/sauce layers `B`: according to the instructions white wine should be served when `A > B`, which means that the `redWine` function should return `true` when `A <= B`, not `A < B`. 

However, in the test case `A == B`, which means that the test fails if a student implements the `redWine` function to return `true` when `A <= B`. To fix this, I added an extra `"mozzarella"` layer to the test case, so that it represents the case `A > B` and therefore correctly expects `false`.

### DIFF
--- a/exercises/concept/lasagna-master/.meta/Sources/LasagnaMaster/LasagnaMasterExemplar.swift
+++ b/exercises/concept/lasagna-master/.meta/Sources/LasagnaMaster/LasagnaMasterExemplar.swift
@@ -60,6 +60,6 @@ func redWine(layers: String...) -> Bool {
     return count
   }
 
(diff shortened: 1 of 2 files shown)


SleeplessByte:
cc @ErikSchierboom

SleeplessByte:
(Haha hoi @sanderploegsma 👋🏽)

ErikSchierboom (approved):
Thanks!

PR #507, opened 2021-11-04
Changed expected value for 'testForcefulQuestions' unit test

Ryan-Gross1993:
The input: "WHAT THE HELL WERE YOU THINKING?"

It's a question that is in all caps, so, it meets the criteria for yelling a question.

Current expected value does not match what should be returned when a question is yelled.

### DIFF
--- a/exercises/practice/bob/Tests/BobTests/BobTests.swift
+++ b/exercises/practice/bob/Tests/BobTests/BobTests.swift
@@ -40,7 +40,7 @@ class BobTests: XCTestCase {
 
     func testForcefulQuestions() {
         let input = "WHAT THE HELL WERE YOU THINKING?"
-        let expected = "Whoa, chill out!"
+        let expected = "Calm down, I know what I'm doing!"
         let result = Bob.hey(input)
         XCTAssertEqual(expected, result)
     }


kotp:
According to the description, there is no specification for a yelled question:

```
# Description

Bob is a lackadaisical teenager. In conversation, his responses are very limited.

Bob answers 'Sure.' if you ask him a question.

He answers 'Whoa, chill out!' if you yell at him.

He says 'Fine. Be that way!' if you address him without actually saying
anything.

He answers 'Whatever.' to anything else.
```

So if this test comes in, perhaps the description file should be updated as well.

If this is something that is really missing and should be addressed because it improves the exercise, then perhaps an issue should be opened in [problem-specifications][1].

[1]: https://github.com/exercism/problem-specifications/issues

ErikSchierboom:
I think the current implementation is based on an older version of the spec. The latest version includes the question/shout combination you are referrring to: https://github.com/exercism/problem-specifications/blob/main/exercises/bob/canonical-data.json Would you perhaps be willing to update the exercise to have it include all the latest tests?

Ryan-Gross1993:
I can update it.

For next time, if I notice an issue with any unit tests, I should go through problem-specifications? 

I found a syntax error in another problem. :p

ErikSchierboom:
@Ryan-Gross1993 It depends. It never hurts to check problem-specifications to see if there is also an error in the source, but in general that will likely not be the case.

bwcDvorak:
As a learner, I got hung up on this for a few days looking for mistakes in my code, then pulled down the Bob lesson on another language track and realized the test itself appears to be the problem, then ultimately found this issue filed here. Just going to correct the test file locally for now..

kotp:
For the language track that you are using, I would not change the test locally, except for attending to the skip directives, if any.  They are designed to be the way they are, and hopefully not contradictory to the README.md that is delivered to you.  Instead, change your solution, not the tests locally.

If the tests need to be changed, then they need to be changed for everyone on the track, rather than for only one student.

kotp:
To be clear, yes, if you want to the suggested changes here, locally, I think that is probably OK, you are probably getting an update that will be in at some point.  But be careful to not submit that test, as it will prevent you from having updates able to be done automatically for you in the future.

Instead, you might treat the exercise as a "bug fix" and deprecate the earlier behavior once the tests are updated.

bwcDvorak (approved):
README.md for this lesson includes 

`He answers 'Whoa, chill out!' if you YELL AT HIM (in all capitals).
He answers 'Calm down, I know what I'm doing!' if you yell a question at him.`

Since the input string is all caps and ends in a question mark, the "Calm down" response appears most appropriate. 

("Chill out" is the response for yelled statements.)

bwcDvorak:
Don't want to break things further, so have backed out my local changes - thanks for the assistance!

kotp:
If the calm down is appropriate for the response for a yelled question, then this should be opened up as an issue (if it is not already there) for https://github.com/exercism/problem-specifications and that additional test or test correction will be availble for all tracks that choose to include the exercise.

kotp:
I will say that often the README is (perhaps purposefully) not completely specific.  The purpose of the tests is to give some clarity to anything that may be ambiguous in the README, or leave it up to the student and mentor to flesh out as a learning experiment.

PR #509, opened 2021-11-11
Improved bad example in high-score-board introduction

raxavelo:
var emptyDict2 = ["Character": "String"]() generates an error.
Correct declaration of empty dictionary is:
var emptyDict2 = [Character: String]()


ErikSchierboom (approved):
Thanks!

PR #511, opened 2021-11-12
Correction in first exercise

raxavelo:
Instruction "The initial width and height should be 60 and 80, respectively." is wrong. It should be: "80 and 60, respectively." Users are loosing a lot of time because that error in instruction.


ErikSchierboom (approved):
Good catch

PR #512, opened 2021-11-13
[Characters Concept] Fix typo with full stop

mathstrains21:
There is an extra full stop in the second one, which led me to believe that that comparison was correct. I researched it and it is not. Here is a fix for it!


ErikSchierboom (approved):
Thanks!

Issue #476, opened 2021-08-24
Fix CI


ErikSchierboom:
This repository's `main` branch currently does not pass all CI checks.
This makes detecting issues in PRs harder and might result in a broken track, and so is an important thing to fix.

The following CI workflows are currently failing:

https://github.com/exercism/swift/runs/3370849643

## Tracking

https://github.com/exercism/v3-launch/issues/55

sanderploegsma:
Should be fixed after merging #484 and #487 I think

kerolloz:
@ErikSchierboom 
Tests pass here 
![image](https://user-images.githubusercontent.com/36763164/196004963-1edab6da-6d9b-4688-9ead-4670226494da.png)

PR #520, opened 2021-11-23
Fix ci linter issue for missing slug

wsaults:
This resolves the following issue:

```
`/home/runner/work/swift/swift/concepts` contains a directory named `constants-and-variables`, which is not a `slug` in the concepts array. Please add the concept to that array:
/home/runner/work/swift/swift/config.json
```

Can be seen here: https://github.com/exercism/swift/runs/4304703883?check_suite_focus=true

This may also relate to issue: https://github.com/exercism/swift/issues/476

### DIFF
--- a/config.json
+++ b/config.json
@@ -1054,6 +1054,11 @@
       "name": "Constants"
     },
     {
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
Nice!

PR #529, opened 2021-12-31
Update BobTests.swift

kalafus:
Specification states, "He answers 'Calm down, I know what I'm doing!' if you yell a question at him."


ErikSchierboom (commented):
Good catch! Could you also update the example implementation at https://github.com/exercism/swift/blob/main/exercises/practice/bob/.meta/Sources/Bob/BobExample.swift?

PR #530, opened 2021-12-31
Update GradeSchoolTests.swift

kalafus:
fix test blocking error: "warning: variable 'school' was never mutated; consider changing to 'let' constant"


ErikSchierboom:
If I look at the [CI run](https://github.com/exercism/swift/runs/4672036687?check_suite_focus=true), the warnings are now errors:

```console
error: cannot use mutating member on immutable value: 'school' is a 'let' constant
```

Should only one tests have been changed perhaps?

kytrinyx:
As we've not heard back from the original author here, and the tests are failing, I'm going to go ahead and close this.

PR #534, opened 2022-01-26
fix instruction 4 parameter type

jancassio:
fix text at step 4 description where is expected a `Int` type for `year` parameter instead of `string` type.


ErikSchierboom (approved):
Thanks!

Issue #541, opened 2022-03-21
Some concept exercises have 2 source and test directories


dlwalter:
A few concept exercises (for example, Santa Swifty Helper, and Magician in Training) have a Source directory as well as an org-sources directory.  Same goes for Tests and org-tests directory. I'm guessing this is leftover from an update?

This is confusing for new users.

ErikSchierboom:
@dlwalter Would you be willing to submit a PR to fix this?

dlwalter:
@ErikSchierboom done.

https://github.com/exercism/swift/pull/542

ErikSchierboom:
Closed by #542

PR #542, opened 2022-03-22
Removed org-tests and org-sources folders since they contain duplicate code

dlwalter:
A few exercise folders contained folders `org-sources` and `org-tests` in addition to the `Source` and `Tests` folders which presents confusion to beginner users.

https://github.com/exercism/swift/issues/541

`org-sources` and `org-tests` have been removed from `magician-in-training` and `santas-helper`


ErikSchierboom (approved):
Thanks!

PR #545, opened 2022-04-02
Fixed typo in dict2 declaration

nikitakaralyus:
```swift
var emptyDict2 = ["Character": "String"]()
```

This declaration will not compile as there is key and value provided, not the types of the Dictionary.

Screenshot 2022-04-02 at 16 26 47

The correct form should look like this:

Screenshot 2022-04-02 at 16 27 59


ErikSchierboom (approved):
Good catch!

PR #546, opened 2022-04-13
Give a little more direction

Atomic-Germ:
Give the learner a little more direction as to what's being tested.


ErikSchierboom (commented):
The `instructions.md` are all synced from a common source: https://github.com/exercism/problem-specifications/blob/main/exercises/rna-transcription/description.md

I can see that this file has become outdated, so you could consider syncing it to the latest version.

That said, we don't allow directly changing this file. What you could do, is to create an `instructions.append.md` with the Swift-specific instructions.

PR #551, opened 2022-04-16
Explicit expectations

Atomic-Germ:
This will help to avoid needing to inspect the tests in order to understand the expected names and structures. It should be edited by a maintainer to better fit if needed.

### DIFF
--- a/exercises/practice/gigasecond/Sources/Gigasecond/Gigasecond.swift
+++ b/exercises/practice/gigasecond/Sources/Gigasecond/Gigasecond.swift
@@ -1,1 +1,2 @@
-//Solution goes in Sources
+// Define "Gigasecond" with a "description" property containing
+// the result, times are expected in "yyyy-MM-dd'T'HH:mm:ss" format


ErikSchierboom (commented):
I think the Swift track doesn't do this currently, but many tracks provide stubs in their solution files. This is in fact our recommended approach for compiled languages: https://exercism.org/docs/building/tracks/practice-exercises#h-file-stub-implementation

Atomic-Germ:
It would absolutely make sense to be doing this in the Swift track. The worst of the current way for me is needing to look at the test files just to know what to name various objects in the code. The least is just how much `//Solution goes in Sources` seems unfinished. I'll go back to this one and see how well I can fit it to the other compiled languages.

ErikSchierboom:
> I'll go back to this one and see how well I can fit it to the other compiled languages.

Here is a C# example: https://github.com/exercism/csharp/blob/main/exercises/practice/acronym/Acronym.cs

But dynamic languages sometimes also do this. See this Clojure example: https://github.com/exercism/clojure/blob/main/exercises/practice/anagram/src/anagram.clj

kytrinyx:
I'm going to go ahead and merge this, and we can take a look at making stubs for all the exercises in a second pass.

kytrinyx (approved):

PR #557, opened 2022-05-04
Fix expected value for testTotalMinutes

Jokushu:
Following the description, it should be testRemainingMinutes(elapsedMinutes: 13) + testPreparationMinutes(layers: 6) which return 27 and 12 and the total is 39, not 25

### DIFF
--- a/exercises/concept/lasagna/Tests/LasagnaTests/LasagnaTests.swift
+++ b/exercises/concept/lasagna/Tests/LasagnaTests/LasagnaTests.swift
@@ -20,7 +20,7 @@ final class LasagnaTests: XCTestCase {
 
   func testTotalMinutes() throws {
     try XCTSkipIf(true && !runAll)  // change true to false to run this test
-    XCTAssertEqual(totalTimeInMinutes(layers: 6, elapsedMinutes: 13), 25)
+    XCTAssertEqual(totalTimeInMinutes(layers: 6, elapsedMinutes: 13), 39)
   }
 
   static var allTests = [


ErikSchierboom:
I think you might have misunderstood the description of the task. You are meant to return the time that has already been spent, which does not include the remaining minutes. The correct time is therefore: `elapsedMinutes + testPreparationMinutes(layers: 6)`

Jokushu:
Oh sorry for this... I missed on this 

> The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.

Thank you for the explaination

Jokushu:
> I think you might have misunderstood the description of the task. You are meant to return the time that has already been spent, which does not include the remaining minutes. The correct time is therefore: `elapsedMinutes + testPreparationMinutes(layers: 6)`

Oh sorry for this... I missed on this 

```
 The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
```
Thank you for the explanation

ErikSchierboom:
Sure!

nacho4d:
Hi,

may I suggest to change function name from `totalTimeInMinutes` to `totalSpentTimeInMinutes` ?

I read only a part of the problem and I thought it was 39... I know it is my bad since I didn't read until the end. However, this can be prevented a bit by renaming the function ;-]

PR #560, opened 2022-05-06
Add percent symbol

Will-1-Am:
Makes instructions more readable and understandable.

### DIFF
--- a/exercises/concept/vehicle-purchase/.docs/instructions.md
+++ b/exercises/concept/vehicle-purchase/.docs/instructions.md
@@ -6,7 +6,7 @@ You have three tasks, one to help you determine the price of the vehicle you can
 
 ## 1. Compute whether or not you can afford the monthly payments on a given car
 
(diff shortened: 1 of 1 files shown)


Jokushu:
I think that `within 10 of your monthly budget` means `monthlyBudget-10` or `monthlyBudget+10`.
If your monthly budget is 100 and the monthly payment is in the range of 90...110, it returns true as we want -10 and +10 to behave the same way.

Will-1-Am:
I totally agree that the instructions could be interpreted exactly as you state, but on
retrospection, it doesn't really make sense to interpret them in that manner.

Considering the example calls for Task 1, the first call is for a very inexpensive car 
fitting a very limited budget.  The second example call is for a car costing over 2 million 
fitting an entirely different budget - ten dollars either way on this kind of budget 
really becomes almost meaningless.  

A budget tolerance of 10%, on the other hand, on a budget of 10,000 (e.g. a max of 11,000) 
is probably a reasonably consideration for someone of such means. Compare that to a budget
tolerance of 10% on a budget of 100 (e.g. a max of 110). Someone with this lower budget will 
be more sensitive to a change of this magnitude. 

If a purchase is within the budget there is really no reason to be frugal, making the 
(budget - 10) value kind of pointless.  The budget + 10 does have real meaning since frugality 
could mean the difference of acquiring the asset or not.

Just writing this explanation, highlights the grey areas involved by not including a
symbol whether it be a % or a currency symbol ¢, $, or £, etc.  It might be argued
that by the term budget, the author meant income, although these two should
never be confused.  IMHO, the percent symbol makes the most sense to add to the 
instructions as it bears a close relation to a pragmatic budget and code written to this 
specification will still pass the tests.

PR #567, opened 2022-06-15
Clarify task 1 of vehicle-purchase

blynx:
If I'm not mistaken, that sentence gives a wrong clue: `if the monthly payment is within 10% of your monthly budget you will want to return a special reminder to be frugal`. For me it reads that the payment may be between 0 and 10 of a 100 budget, but it is supposed to be 100 - 110.

I hope the correction is proper English: `if the monthly payment is within 10% above your monthly budget you will want to return a special reminder to be frugal`  
(Non native speaker, having a cold, brain slightly molten)


ErikSchierboom (approved):
Nice catch!

PR #568, opened 2022-07-01
Add "and" for numbers over 99

dmytriigolovanov:
Added "and" part for test cases, which number is over 99, according to exercise instructions (step 4).
These numbers must be in the format "... and ...", as in the examples in the instructions.


ErikSchierboom:
Looking at [the instructions](https://github.com/exercism/swift/blob/main/exercises/practice/say/.docs/instructions.md), it looks like the "and" part is an optional thing that you _can_ do. Step 4 does use the existing format though.

kytrinyx:
I'm going to go ahead and close this, as the test suite does follow the spec.

PR #573, opened 2022-08-05
Update HighScoreBoardTests

Jokushu:
Change `updateScore` test value so it equals to expected result
Fix error from previous changes:
```
XCTAssertEqual failed: ("["Amil PAstorius": 99373, "Min-seo Shin": 1999, "Jesse Johnson": 1337]") is not equal to ("["Amil PAstorius": 99373, "Min-seo Shin": 1999, "Jesse Johnson": 2674]") - Updating a player's score should add the update to their score and leave the rest unchanged.
```


Jokushu:
Wrongly read the function decription

PR #575, opened 2022-09-05
Update instructions.md for Clock

a-co:


ErikSchierboom (commented):
Good idea. Unfortunately, you can't just update the `instructions.md` document,as its contents are synced with contents in the exercism/problem-specifications repo. The way to add additional content to the description is to add a `.docs/instructions.append.md` file to the exercise's directory. Like https://github.com/exercism/swift/blob/main/exercises/practice/hello-world/.docs/instructions.append.md

kytrinyx:
Thanks! The docs were synced with problem-specifications in #586. Closing.

PR #497, opened 2021-10-07
Update instructions.md

Aurisimo:
It seems id bites should be started reading from right to left (from rightmost bit).

Issue #496, opened 2021-10-04
BombDefuser exemplar solution and tests


adauguet:
I am trying a to mentor a student on the BombDefuser exercise.

It states:
> For each bit in the ID number, starting with the leftmost bit, you will apply the flipper closure to the wires tuple if the bit is a 0 and you will apply the rotator closure if it is a 1 giving the new state of the wires.

To my understanding, this means that I should process the leftmost bit, which the left most `1` or `0`.
If my number is `253` (like in the tests), its binary representation is `1111 1101` and I should process bits in the following order: `1`, `1`, `1`, `1`, `1`, `1`, `0` and `1`.

In the exemplar solution, the implementation performs: `if bits.isMultiple(of: 2)`. But that is checking the rightmost bit, not the leftmost.

What am I missing? Is this an error in the implementation? Or in the instructions?

Fun fact, the `113` number (used in tests) is working in both directions... leading to some confusion debugging all this.

ErikSchierboom:
These are the [exercise's prerequisites](https://github.com/exercism/swift/blob/main/config.json#L39):

- "functions", "higher-order-functions", "loops"

With just those prerequisites (and their transitive prerequisites), I don't think we can expect the user to know how to do bitwise manipulation. Similarly, the exercise uses the `UInt8` type, which also is not known to the student. I would suggest one of the following:

Option 1:

- Add `arrays` as a prerequisite
- Change the `UInt8` parameter to an array of `Int` values, effectively replacing the single byte with an array of `0` and `1`s

Option 2:

- Add `characters` as a prerequisite
- Change the `UInt8` parameter to a `String` that contains `0` and `1` characters.

adauguet:
You are right, bitwise operators are too complicated for a beginner. Actually there are only explained in one the last chapter of the Apple Swift book.

Are you suggesting to completely replace `Int` by an array `Int`, or a `String`?

ErikSchierboom:
I am.

adauguet:
OK great. Will work on the PR then.

mkalmes:
I've just completed the exercise myself and was confused by the description to start with the leftmost bit. Checking the bits in Calculator.app gave me the clue to my misunderstanding. If you visualize the binary representation as an array, the first bit starts at position 0 and grows to the right. However the string representation shows the Least Significant Bit (LSB) on the *rightmost* side.

[...]

PR #493, opened 2021-10-05
Update instructions for bit enumeration

paiv:
"leftmost bit" is the most significant bit, whereas the exercise is designed for least significant bit.

https://github.com/exercism/swift/blob/91b329ac91bbbe5ac1b670c1e653c70af0d4c389/exercises/concept/bomb-defuser/.meta/Sources/BombDefuser/BombDefuserExemplar.swift#L18-L25

- https://en.wikipedia.org/wiki/Bit_numbering

Upd: just have seen this discussed in #496

PR #578, opened 2022-10-11
bit checking order is reversed in tests

BaseCase:
I believe the tests for this exercise are written expecting we check the bits from right to left, not left to right as indicated here. `testShuffle1` and `testShuffle3` both pass regardless of which direction we go, but `testShuffle2` fails when going left to right.

If starting with leftmost is the correct behavior, then I believe `testShuffle2` should expect to get `("Purple", "Marigold", "Cyan")` instead of the `("Marigold", "Cyan", "Purple")` it currently wants (I could do a different PR with that change instead of this if that's the right thing).

I might also just be confused and it's correct as is!!


ErikSchierboom:
Pinging @exercism/reviewers

paiv:
Linking
- #496
- #493
- #497

BaseCase:
Ah! My mistake, I failed to check existing PRs before opening this one. Apologies for the noise. @paiv, would you like me to close this?

kotp:
This is an identical proposal as #497 and so I am closing as a duplicate.
A discussion is availble in issue #496.

PR #581, opened 2022-10-20
fixed typo

safety-panda:
fixed a typo in the description
"she" => "such"


ErikSchierboom (approved):
If you feel like it, you could also put the keywords (`return`, `continue`, etc.) in backticks. Could also be done in a follow-up PR.

safety-panda:
That's a great suggestion, thanks, I will do.

PR #623, opened 2023-06-14
Fix largest-series-products tests

kytrinyx:
We deleted a couple of tests from the test suite
but they were still being referenced from allTests.

This deletes the stray references.


meatball133:
I think the old test runner has a reference to the allTests.
But the new one removed that reference, so that constant is not really needed, so it could be omitted.

ErikSchierboom (approved):

kytrinyx:
That would explain why CI passed 😅

meatball133:
I will be honest, I am not sure. But I think there were some kind of requirement to have it and then Apple made so you didn't anymore, I think this pr added it for linux: https://github.com/apple/swift-package-manager/pull/2174

Here can you see one of apples repo updating its tests files: https://github.com/apple/swift-syntax/pull/171/files

PR #631, opened 2023-06-19
Remove allTests constant in practice exercises

kytrinyx:
Apple appears to have changed things so that you no longer have to explicitly tell the test runner which tests exist.

This removes the explicitly defined allTests constant from the practice exercises.

~Note that I'm not actually sure this will work—I'm making the PR so that we can see if CI passes.~ _Edit: seems like it's working..._


kytrinyx:
Let me know what you all think. If this looks good I can try doing the same thing for the concept exercises.

meatball133 (approved):
I think it looks good, and it should work.

One thing which concerns me is that this pr would likely cause the website to rerun tests for all exercises on the track. This includes I think around 180.000 iteratons, but I think that should still be fine.

kytrinyx:
Yes, it would definitely cause everything to be re-run. Let's wait for @ErikSchierboom to weigh in on that. We may want to wait to merge to make sure there isn't anything else going on at the same time.

ErikSchierboom:
> I think it looks good, and it should work.

Is there a corresponding test runner PR that verifies this? As in: a PR that updates the golden tests?

ErikSchierboom:
> Yes, it would definitely cause everything to be re-run. 

We'll be working on something to bypass this (not having to run everything).

meatball133:
> Is there a corresponding test runner PR that verifies this? As in: a PR that updates the golden tests?

I could start writing one, but I have done testing which points at that there shouldn't be any problems, both on the test runner and my local machine.

meatball133:
https://github.com/exercism/swift-test-runner/pull/47

PR #633, opened 2023-06-24
fix use of assignment in guard clause

hoelzlmanuel:
See https://forum.exercism.org/t/pr-to-fix-typo-in-vexing-vehicle-purchase-exercise-in-swift-track/6225/2.


meatball133 (approved):
Nice! ✨

PR #634, opened 2023-06-28
Swift "Control transfer" typo fix

nathanscain:


meatball133:
Hi!, it says I can't reopen this pr, since the original repo doesn't exist.

If you like you could open a new pr since this example is incorrect, but I think something like this could make a bit more sense:

```swift
count = 1
repeat {
  count += 1
  if count == 4 { continue }
  print(count)
} while count < 6
```
prints:
```
2
3
5
6
```
rather than:
```
2
3
4
5
6
```

Issue #448, opened 2021-04-16
Use same swift version for concept and practice exercisesl


ErikSchierboom:
The Concept Exercises and Practice Exercises use different versions of swift: 5.3 and 4.2 respectively. This will be confusing to the student and problematic for the test runner. The Practice Exercises should be updated to 5.3.

PR #638, opened 2023-07-09
Update practice exercises to Swift 5.3

ErikSchierboom:
Closes https://github.com/exercism/swift/issues/448


meatball133 (approved):
Just keep in mind that this is a hard limit, using a swift version older than the set one will throw an error. Swift 5.3 is rather old, but just to keep that in mind. You can manually change it to make it compatible with older versions.

PR #692, opened 2023-09-07
Update exercises batch 20

meatball133:
This includes:

- Ocr Numbers
- Phone number


exercises/practice/ocr-numbers/Sources/OcrNumbers/OcrNumbers.swift
```diff
@@ -1 +1,53 @@
-//Solution goes in Sources
+enum OcrNumberError: Error {
```
ErikSchierboom, line 1: I think you want this to be an empty(-ish) stub

ErikSchierboom (approved):

PR #702, opened 2023-10-29
LasagnaTests.swift add missing parentheses

vadimpiven:
I was unable to pass this task due to error
```
/mnt/exercism-iteration/Tests/LasagnaTests/LasagnaTests.swift:9:20: error: add () to forward @autoclosure parameter
    XCTAssertEqual(expectedMinutesInOven, 40)
                   ^~~~~~~~~~~~~~~~~~~~~
                                        ()
error: fatalError
```


meatball133:
The goal of this exercise is to define a constant not a function.

Hence the instruction:

````
## 1. Define the expected oven time in minutes

Define the `expectedMinutesInOven` constant, that stores how many minutes the lasagna should be in the oven.
According to the cooking book, the expected oven time in minutes is 40:

```swift
expectedMinutesInOven
// Returns 40
```
````

PR #708, opened 2023-11-19
[New Exercise]: Armstrong Numbers

meatball133:
This exercise will test large numbers, seems like it could add a lot of complexity for this simple exercise without any real good language support for large numbers. I am considering adding this library: https://github.com/mkrd/Swift-BigInt, which would be the first library added. Since Swift doesnt have native Int128 nor bigint

### DIFF
--- a/config.json
+++ b/config.json
@@ -326,6 +326,14 @@
         "topics": ["maps", "transforming"]
       },
       {
+        "slug": "armstrong-numbers",
+        "name": "Armstrong Numbers",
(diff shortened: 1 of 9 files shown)


IsaacG (approved):

IsaacG:
Should this be synced/updated to match the latest problem specs?

meatball133:
Yeah, it should I will look into that

meatball133:
Looked into it, it looks to be update to date

PR #739, opened 2024-04-12
style: update missing link to Swift documentation reference in Bomb Defuser exercise

mmmmmob:


meatball133:
The current link just redirects the user, could you update to the redirected url instead?

mmmmmob:
@meatball133 Does this url I just committed correct?

meatball133 (approved):
Yes!

PR #759, opened 2024-06-13
Update LasagnaTests.swift

Endington543:
Fixed incorrect function call

Added parenthesis to the expectedMinutesInOven function call in LasangaTests.swift


Endington543:
The incorrect function call causes the

PR #763, opened 2024-08-06
Update introduction.md

csprasad:
Removed Unused Declaration of Variable `speed`.

I noticed that the variable speed is declared but never used in the code example. To avoid confusion, I suggest removing this unused declaration.


BNAndras:
Seems like `speed` is used later in `myChar.stats = CharacterStats(health: 72.8, speed: 19, strength: 6)`. So you'll need to edit that bit later on as well.

meatball133:
I don't really see how this adds confusion, since even though if not used it can still be important for the codebase itself, thereby I wouldnt strictly speaking say that it should be removed

PR #768, opened 2024-09-10
Add Knapsack exercise

kahgoh:
This is one of the [48in24](https://forum.exercism.org/t/48in24-exercise-08-13-knapsack/12054) exercises.

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,9 +8,6 @@
     "representer": false,
     "analyzer": false
   },
-  "test_runner": {
-    "average_run_time": 9
-  },
   "blurb": "Swift is a modern open-source language with the aim of making programming simple things easy, and difficult things possible while still being fast and safe by design. Swift can run in scripting mode in the included REPL or it can be compiled with full support for C and Objective-C interoperability.",
   "version": 3,
   "online_editor": {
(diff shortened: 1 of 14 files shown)


config.json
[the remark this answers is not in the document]
kahgoh: @IsaacG I've reverted the formatting changes from this PR. I'll be happy to fix the formatting in a follow up PR, after this one.

IsaacG:
+ @exercism/swift for review.

exercises/practice/knapsack/.meta/Sources/Knapsack/Item.swift
```diff
@@ -0,0 +1,7 @@
+// Represents an item in knapsack.
+// There is be no need to edit this file and changes to this file are ignored by the Exercism test runners.
+
+struct Item {
```
meatball133, line 4: The indent should be 2 not 4 spaces. Otherwise lgtm

kahgoh: I've just updated the exercise source files to use 2 spaces for indents. I wasn't sure about changing the indenting in the test template though. The test templates in the other exercises seem to use 4 spaces, although the formatter fixes the generated test to use 2 spaces. Let me know if the test template should also use 2 spaces.

IsaacG: I'm not part of the Swift team but it sounds like the templates might benefit from an update. If they do get an update, that should probably be a different PR.

meatball133 (approved):

exercises/practice/knapsack/.meta/Sources/Knapsack/Item.swift
meatball133: The track was a bit of a mess when I took over it tbh, the concepts used 2 indent while the practice exercises used 4 (the online editor used 4). Swift doesn't seem to agree on the indent size. The Swift book (written by apple) uses 4 indent, but if you look at libraries apple writes they use 2 (an example for reference [here](https://github.com/swiftlang/swift-testing/blob/main/Sources/Testing/Traits/Bug.swift)), I also belive the formatted by default uses 2. Thereby there could have been a transition error when I decided to transition the track to 2 indent. You are free to update the templates files but I don't see it as a necessity

meatball133: Or actually, Swift 6 includes a new testing framework so that means that we should likely transition to that at some point which means re-writting all test files so likely the easiest thing would be to just apply the fix then.

PR #801, opened 2024-12-21
[Swift 6]: Update Exercises batch 17

meatball133:
Includes:
- Space-age
- Strain
- Sublist
- Sum-of-multiples

Noteworthy changes:

Addition of the swift library: https://github.com/apple/swift-numerics, this is due to that the new testing library has no proper way of comparing floating point numbers,

> The testing library doesn’t provide an equivalent of [XCTAssertEqual(_:_:accuracy:_:file:line:)](https://developer.apple.com/documentation/xctest/3551607-xctassertequal). To compare two numeric values within a specified accuracy, use isApproximatelyEqual() from [swift-numerics](https://github.com/apple/swift-numerics).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -53,7 +53,8 @@ let practiceExerciseTargets: [Target] = practiceExercises.flatMap {
         .testTarget(
             name:"\($0.pascalCased)Tests", 
             dependencies: [
-                .target(name:"\($0.pascalCased)")
+                .target(name:"\($0.pascalCased)"),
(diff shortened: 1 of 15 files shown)


IsaacG (approved):

IsaacG:
> i was updating Space Age on my swift track and i noticed that `SpaceAge.swift` in the `Sources/` directory was the standard blank exercise-starter file, but it now contains a solution to the exercise.

Thanks for catching that! The best place to raise these issues would be on the forum under Programming > Swift

PR #820, opened 2024-12-30
Update vehicle purchase instructions to be more clear

### LINKED ISSUE #810: Errors in conditionals/vehicle-purchase?
Going over the conditionals exercise (vehicle-purchase), it has these two tests:

https://github.com/exercism/swift/blob/4dea7e2b6dc717d04c1e68cb1221dd33f6416df5/exercises/concept/vehicle-purchase/Tests/VehiclePurchaseTests/VehiclePurchaseTests.swift#L9-L11

https://github.com/exercism/swift/blob/4dea7e2b6dc717d04c1e68cb1221dd33f6416df5/exercises/concept/vehicle-purchase/Tests/VehiclePurchaseTests/VehiclePurchaseTests.swift#L23-L25

But the instructions say:

> The function should return the following message based on the following conditions:
> 
> - If the price of the vehicle is less than or equal to the monthly budget, return the message "Yes! I'm getting a &lt;vehicle>".
> - If the price of the vehicle is 10% above your monthly budget, return the message "I'll have to be frugal if I want a &lt;vehicle>".
> - If the price of the vehicle is more than 10% above your monthly budget, return the message "Darn! No &lt;vehicle> for me".

516.32 is more than 10% above the monthly budget (100), so it shouldn't say yes. Likewise, 12,500 is quite a bit more than 10% over the monthly budget of 200, so it shouldn't say you have to be frugal.

meatball133:
Resolves #810

### DIFF
--- a/exercises/concept/vehicle-purchase/.docs/instructions.md
+++ b/exercises/concept/vehicle-purchase/.docs/instructions.md
@@ -9,15 +9,17 @@ You have three tasks, one to determine if you will need one to help you choose b
 The auto dealers in your town are all running a five year, 0% interest promotion that you would like to take advantage of.
 But you are not sure if you can afford the monthly payments on the car you want.
 
+The monthly payment is the cars total price divided by the number of months under the five year period.
+
 Implement the `canIBuy(vehicle:price:monthlyBudget:)` function that takes the following arguments:
 - `vehicle` - The name of the vehicle you want to buy.
 - `price` - The price of the vehicle you want to buy.
 - `monthlyBudget` - The amount of money you can afford to pay each month.
 
 The function should return the following message based on the following conditions:
-- If the price of the vehicle is less than or equal to the monthly budget, return the message `"Yes! I'm getting a <vehicle>"`.
-- If the price of the vehicle is 10% above your monthly budget, return the message `"I'll have to be frugal if I want a <vehicle>"`.
-- If the price of the vehicle is more than 10% above your monthly budget, return the message `"Darn! No <vehicle> for me"`.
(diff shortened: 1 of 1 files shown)


exercises/concept/vehicle-purchase/.docs/instructions.md
```diff
@@ -9,15 +9,17 @@ You have three tasks, one to determine if you will need one to help you choose b
...
 - `vehicle` - The name of the vehicle you want to buy.
 - `price` - The price of the vehicle you want to buy.
 - `monthlyBudget` - The amount of money you can afford to pay each month.
 
 The function should return the following message based on the following conditions:
-- If the price of the vehicle is less than or equal to the monthly budget, return the message `"Yes! I'm getting a <vehicle>"`.
-- If the price of the vehicle is 10% above your monthly budget, return the message `"I'll have to be frugal if I want a <vehicle>"`.
-- If the price of the vehicle is more than 10% above your monthly budget, return the message `"Darn! No <vehicle> for me"`.
+- If the monthly payment of the vehicle is less than or equal to the monthly budget, return the message `"Yes! I'm getting a <vehicle>"`.
+- If the monthly payment of the vehicle is within 10% your monthly budget, return the message `"I'll have to be frugal if I want a <vehicle>"`.
```
IsaacG, line 21: "within" can be interpreted as "+/- 10%" is in the range of 90% to 110%. Is that the intent here?

meatball133: Good catch, do you think "up to", would be a better word of choice?

IsaacG: Yes. Maybe "above your budget by up to 10% (inclusive)"

IsaacG (approved):

PR #821, opened 2025-01-02
Update README to reflect Swift version 6.0.3 and testing library changes

meatball133:


README.md
```diff
@@ -17,8 +17,12 @@ Hi. &nbsp;👋🏽 &nbsp;👋 &nbsp;**We are happy you are here.**&nbsp; 🎉&nb
 **`exercism/Swift`** is one of many programming language tracks on [exercism(dot)org][exercism-website].
 This repo holds all the instructions, tests, code, & support files for Swift _exercises_ currently under development or implemented & available for students.
 
-🌟 &nbsp;&nbsp;Track tooling (_test-runner and Continuous Integration_) runs on Swift `5.8.1`. <br>
```
IsaacG, line 20: Why are you inserting `&nbsp` in markdown?

README.md
```diff
@@ -17,8 +17,12 @@ Hi. &nbsp;👋🏽 &nbsp;👋 &nbsp;**We are happy you are here.**&nbsp; 🎉&nb
 **`exercism/Swift`** is one of many programming language tracks on [exercism(dot)org][exercism-website].
 This repo holds all the instructions, tests, code, & support files for Swift _exercises_ currently under development or implemented & available for students.
 
-🌟 &nbsp;&nbsp;Track tooling (_test-runner and Continuous Integration_) runs on Swift `5.8.1`. <br>
-🌟 &nbsp;&nbsp;Most exercises are solvable with Swift `5.3.0` or higher.
+🌟 &nbsp;&nbsp;Track tooling (_test-runner and Continuous Integration_) runs on Swift `6.1.0`. <br>
+🌟 &nbsp;&nbsp;Most exercises are solvable with Swift `6.0.0` or higher.
+
+As of 2025 does the track use the new [swift-testing][swift-testing] library for testing. This makes the so you have to use Swift 6.0.0 or higher to run the tests.
```
IsaacG, line 23:
```suggestion
As of 2025 does the track use the new [swift-testing][swift-testing] library for testing.
This makes the so you have to use Swift 6.0.0 or higher to run the tests.

```

README.md
```diff
@@ -65,7 +69,7 @@ Please keep in mind [Chesterton's Fence][chestertons-fence].
 
 ## Swift Software and Documentation
 
-**Copyright © 2023 Apple Inc. All rights reserved.**
+**Copyright © 2025 Apple Inc. All rights reserved.**
```
IsaacG, line 72: Why is the readme tagged with an Apply copyright?

README.md
BNAndras:
```suggestion
Currently, the track uses the new [swift-testing][swift-testing] library, requiring Swift 6.0.0 or higher to run the tests.
```

README.md
BNAndras: The README appears to be an adaptation of the Python track readme. The `&nbsp` usage comes from there. https://github.com/exercism/python/blob/9b7a74b59eb82a65d926532f4590af3eb6861c30/README.md?plain=1#L39

README.md
BNAndras: The Swift logo is an Apple trademark so it should be cited as such. However, Apple's documentation says to mention "The Swift logo is a trademark of Apple Inc.", so we should include that. The original copyright notice probably was adapted from the PSF copyright notice in the Python track README.

README.md
```diff
@@ -17,8 +17,12 @@ Hi. &nbsp;👋🏽 &nbsp;👋 &nbsp;**We are happy you are here.**&nbsp; 🎉&nb
 **`exercism/Swift`** is one of many programming language tracks on [exercism(dot)org][exercism-website].
 This repo holds all the instructions, tests, code, & support files for Swift _exercises_ currently under development or implemented & available for students.
 
-🌟 &nbsp;&nbsp;Track tooling (_test-runner and Continuous Integration_) runs on Swift `5.8.1`. <br>
-🌟 &nbsp;&nbsp;Most exercises are solvable with Swift `5.3.0` or higher.
+🌟 &nbsp;&nbsp;Track tooling (_test-runner and Continuous Integration_) runs on Swift `6.1.0`. <br>
+🌟 &nbsp;&nbsp;Most exercises are solvable with Swift `6.0.0` or higher.
+
+As of 2025 does the track use the new [swift-testing][swift-testing] library for testing. This makes the so you have to use Swift 6.0.0 or higher to run the tests.
+If you wish to still run the tests with the old XCTest framework, you can do so by pulling the `5.x` branch.
```
BNAndras, line 24: Students can work locally with the old tests, but can they submit their code to the site? The track would run the Swift 6.0 tests so would that break tests on submission? Some more detail here could be useful.

README.md
meatball133: Yes

README.md
meatball133: Not really. Most earlier solutions will still work. I think around 5-10% of the exercises have changed test cases, and those will not exist on the 5.x changes so those will likely not work.

README.md
IsaacG: If the logo is trademarked by Apple, maybe we shouldn't be using it?

meatball133: Legally speaking is the name Swift trademarked as well so not only the logo. Most languages haven't cared about this, if you check most languages on the website which uses the real logo. So does most of them require you to refer to say who owns the logo. However I have a very hard time to see any company doing a dmca takedown anyways.

IsaacG: +cc @iHiD for FYI/thoughts

iHiD: I think we can get rid of this line. I'm happy with the line below.

IsaacG (approved):

PR #831, opened 2025-02-12
[Swift 6]: Reworked tuples and its exercise

meatball133:


concepts/tuples/about.md
```diff
@@ -1,86 +1,109 @@
...
+
+```swift
+let orderItem = order.item
+// Returns "biscuits"
+let orderQuantity = order.quantity
+// Returns 3
 ```
 
-The components of a tuple may also be accessed through _decomposition_:
+Same as non-named tuples, you can specify the type of each value in a named tuple:
```
IsaacG, line 49:
```suggestion
As with non-named tuples, you can specify the type of each value in a named tuple:
```

concepts/tuples/about.md
```diff
@@ -1,86 +1,109 @@
...
+order.2 = 5
+
+order
+// Returns ("cookies", 1.99, 5)
 ```
 
-You can provide more semantic information about a tuple by supplying labels to the components of a tuple. The individual elements can then be accessed by either their index or their label:
+## Multiple return values
+
+Functions in Swift can only return a single value.
```
IsaacG, line 71: Are tuples a single value or a single object (containing multiple values)?

IsaacG (approved):

Slarky:
I'm just starting with Swift but shouldn't functions specify the return type?

So `func updateQuantity(_ items: [(name: String, amount: Int)], toy: String, amount: Int) {` should be `func updateQuantity(_ items: [(name: String, amount: Int)], toy: String, amount: Int) -> [(name: String, amount: Int)] {` and `func addCategory(_ items: [(name: String, amount: Int)], category: String) {` should be `func addCategory(_ items: [(name: String, amount: Int)], category: String) -> [(name: String, amount: Int, category: String)] {` in https://github.com/exercism/swift/blob/main/exercises/concept/santas-helper/Sources/SantasHelper/SantasHelper.swift?

meatball133:
Yes that was a miss, they should have return type.

However, please do not comment on closed prs, either open an issue or open a topic on the forum. If I didn't have notification on, I wouldn't find this.

PR #833, opened 2025-03-03
[Swift 6]: Update Expert mixolgist and its concept

meatball133:
This renames the exercise master mixologist to expert mixologist and also renames "while loops" to "while and repeat loops"


concepts/conditionals-switch/about.md
```diff
@@ -12,11 +32,7 @@ if str == "apple" {
 }
 ```
 
-While an else-if variant of `if` statements, like the one above, cleans things up considerably compared to a series of `if` statements without `else` clauses, there is still a lot of noise in the code from all of the curly braces. This is where the `switch` statement comes into play. In conditional statements with many possible branches, the switch statement shines. Note, however, that `switch` statements do work a bit differently from `if` statements.
-
-Rather than evaluating a Boolean expression and using the value of that expression to choose the code branch that is run, a simple switch statement takes an input value (or expression which it evaluates to obtain the input value) of some type and compares against one or more values of the same type. If a case is found that matches the input value, the corresponding block of code is run.
-
-Note that all possible cases must be covered in a `switch` statement. In cases like the above where all possible strings _cannot_ be enumerated, a `default` case can be used to match all remaining cases.
+Could instead be written as:
```
IsaacG, line 35:
```suggestion
This could instead be written as:
```

concepts/control-transfer/about.md
```diff
@@ -46,97 +40,50 @@ for fruit in ["banana", "grapes", "apple", "strawberry", "kiwi", "lemon"] {
...
-lmn
-***
+// prints:
+// --- banana ---
+// b
+// a
 ```
 
-However, if we label the outer loop, calling it `mainLoop`, and specify continuing _that_ loop when the current `letter` is "e" then any non-vowel letters following the "e" are not printed, nor is the trailing line of asterisks and newline.
+## Fallthrough
```
IsaacG, line 66:
```suggestion
## Fall through
```

IsaacG: If you mean the keyword, it should be lowercase and not titlecase.

concepts/control-transfer/introduction.md
```diff
@@ -1,28 +1,68 @@
 # Introduction
 
-Sometimes it is necessary to skip to the next iteration of a loop early, without completing the rest of the statements in the current iteration of the loop. The `continue` keyword can be used for this. When `continue` is executed, the loop jump to the next check to see if the next iteration of the loop can be run, i.e. the `while` in while and repeat-while loops or the check if there's another element in the sequence in for-in loops.
+Swift has a set of statements that can be used to alter the normal control flow of loops.
+These are known as control transfer statements.
+And you have already seen one of them, `return`, but this concept will introduce the other onces, which iclude `continue`, `break`, and labels.
+And one which we wont cover in this concept, `fallthrough`, but you can read about it in the [Swift book][fallthrough].
```
IsaacG, line 6:
```suggestion
You have already seen `return`; this concept will introduce other, which include `continue`, `break`, and labels.
We won't cover `fallthrough` in this concept, but you can read about it in the [Swift book][fallthrough].
```

concepts/control-transfer/introduction.md
```diff
@@ -1,28 +1,68 @@
...
+
+## continue
```
IsaacG, line 8:
```suggestion
## Continue
```

concepts/control-transfer/introduction.md
```diff
@@ -1,28 +1,68 @@
...
+
+Sometimes it is necessary to skip to the next iteration of a loop early, without completing the rest of the statements in the current iteration of the loop.
+The `continue` keyword can be used for this.
+When `continue` is executed, the loop jump to the next check to see if the next iteration of the loop can be run, i.e. the `while` in while and repeat-while loops or the check if there's another element in the sequence in for-in loops.
```
IsaacG, line 12:
```suggestion
When `continue` is executed, the loop jumps to the next check to see if the next iteration of the loop can be run, i.e. the `while` in while and repeat-while loops or the check if there's another element in the sequence in for-in loops.
```

concepts/while-and-repeat-loops/about.md
```diff
@@ -0,0 +1,71 @@
+# About
+
+A [while loop][while-loops] is a control flow statement that allows code to be executed while a certain condition is true.
+Swift provides two kinds of while loops: while and repeat-while loops.
+While loop's condition is evaluated before the loop body is executed, and if the condition is false, the loop body will not be executed at all.
```
IsaacG, line 5:
```suggestion
While loop's condition is evaluated before the loop body is executed; if the condition is false, the loop body will not be executed at all.
```

concepts/while-and-repeat-loops/about.md
```diff
@@ -0,0 +1,71 @@
...
+repeat-while loops in Swift have the following structure:
+
+```swift
+repeat {
+  <#code#>
+} while <#condition#>
+```
+
+Here the loop will only execute once, since the condition is false.
+But since it is a repeat-while loop, the loop body will execute once before the condition is evaluated.
```
IsaacG, line 41:
```suggestion
Since it is a repeat-while loop, the loop body will execute once before the condition is evaluated.
```

concepts/conditionals-switch/about.md
```diff
@@ -1,5 +1,25 @@
 # About
 
+A [switch statement][switch] allows you to compare a value against multiple possible matching values.
+It allows you to write more concise code than using multiple `else-if` statements.
+Switch statement starts with the `switch` keyword followed by the value to be compared.
```
BNAndras, line 5:
```suggestion
A switch statement starts with the `switch` keyword followed by the value to be compared.
```

concepts/conditionals-switch/about.md
```diff
@@ -1,5 +1,25 @@
...
+Then do define the cases using the `case` keyword followed by the value to be compared against.
```
BNAndras, line 6:
```suggestion
Then you can define the cases using the `case` keyword followed by the value to be compared against.
```

concepts/conditionals-switch/about.md
```diff
@@ -1,5 +1,25 @@
...
+The `default` keyword is used to define the default case when none of the cases match.
+Like `else-if` statements, `switch` statements can execute the first matching case and then exit the switch statement.
```
BNAndras, line 8:
```suggestion
Like `else-if` statements, `switch` statements only execute the block of code associated with the first matching case.
```

concepts/conditionals-switch/about.md
```diff
@@ -31,13 +47,10 @@ default:
...
 
-Additionally, note that if multiple cases of a switch statement match the input value, only the first matching case is used. An underscore (`_`) can be used to match all values.
-
-Unlike in some other languages, `switch` cases in Swift do not fall through to the next case unless that behavior is explicitly called for with the `fallthrough` keyword. this is the opposite behavior from C which requires explicit `break` statements to prevent fallthrough.
-
 ## Binding and where statements
 
-The values being matched in `switch` statements can also be bound to names which can be used in the body of the case. They can also be used in `where` clauses, which are additional boolean expressions that must evaluate as `true` for the case to match.
+The values being matched in `switch` statements can also be bound to names which can be used in the body of the case.
+They can also be used in `where` clauses, which are additional boolean expressions that must evaluate as `true` for the case to match.
```
BNAndras, line 53:
```suggestion
They can also be used in `where` clauses, which are additional Boolean expressions that must evaluate as `true` for the case to match.
```

concepts/control-transfer/about.md
```diff
@@ -46,97 +40,50 @@ for fruit in ["banana", "grapes", "apple", "strawberry", "kiwi", "lemon"] {
...
-    } else {
-      print(letter, terminator: "")
-    }
-  }
-}
-```
-
-The output of this version of the loop is:
+The `fallthrough` keyword is used to fall through to the next case in a switch statement.
+Meaning that the code in the next case will be executed, even if the value in the next case does not match the value being compared.
```
BNAndras, line 69:
```suggestion
This means the the code within the next case will be executed without checking its condition(s).
```

config.json
```diff
@@ -145,15 +145,13 @@
         "status": "active"
       },
       {
-        "slug": "master-mixologist",
-        "name": "Master Mixologist",
+        "slug": "expert-mixologist",
```
BNAndras, line 148: Is it okay to change the slug? Would Jeremy need to change something in the database to link the old and new slugs?

meatball133: https://discord.com/channels/854117591135027261/1341353705668218900/1341406222439682069

concepts/control-transfer/about.md
```diff
@@ -1,39 +1,34 @@
-# About
+# Introduction
 
-The normal control flow of loops in Swift can be altered using [Swift's control transfer keywords][control-transfer]. One of these keywords, `return` has been seen before in the basics concept exercise. With the use of `return`, if the loop is inside a function, the function will exit at that point, returning whatever value is specified, just as it would at any other point in a function. Two more control transfer keywords that are often used with loops are `continue` and `break`.
+Swift has a set of statements that can be used to alter the normal control flow of loops.
+These are known as control transfer statements.
+You have already seen `return`; this concept will introduce other, which include `continue`, `break`, and labels.
+We won't cover `fallthrough` in this concept, but you can read about it in the [Swift book][fallthrough].
```
IsaacG, line 6: Is Fallthrough covered? I see it below.

concepts/while-and-repeat-loops/about.md
```diff
@@ -0,0 +1,71 @@
+# About
+
+A [while loop][while-loops] is a control flow statement that allows code to be executed while a certain condition is true.
+Swift provides two kinds of while loops: while and repeat-while loops.
+While loop's condition is evaluated before the loop body is executed; if the condition is false, the loop body will not be executed at all.
+while-loops in Swift have the following structure:
```
IsaacG, line 6: One sentence starts with "While loop's" and the next with "while-loops". It would be good to make these consistent.

IsaacG (approved):

concepts/control-transfer/about.md
meatball133: The introduction doesn't cover it to keep the text concise but you are right, that shouldn't be in the about docs.

PR #837, opened 2025-03-27
Update instructions.append.md to remove variable in function call.

erinn:
A test case for using a variable in the function call doesn't exist so it seems that the hint should probably just be about what is actually tested for. The other option I suppose is to put a test case in for that other eventuality, but I am not savvy enough to do that at this point.


IsaacG:
@meatball133  Do we need the instructions.append file at all?

meatball133:
This is already a planned change: https://github.com/exercism/swift/pull/791, yes this pr has been laying for like a few months at this point. However, the Swift 6.0 update is imminent, Swift 6.1 _should_ be releasing next week, I need perhaps a week or two after then and then we should be able to merge in the content.

tl;dr: This will be fix in like a month but as of now it will just be as it is.

PR #849, opened 2025-06-04
fix: birdwatcher template and document issues

MojtabaHs:
All changes are fully backward compatible.

So we can add `[no important files changed]` to merge-commit message.

### DIFF
--- a/concepts/for-loops/about.md
+++ b/concepts/for-loops/about.md
@@ -9,7 +9,7 @@ Swift allows you to iterate over a variety of sequences, such as ranges, arrays,
 When every element of the sequence has been iterated over, the loop exits.
 
 For loops are declared by using the `for` keyword, followed by a variable name, the `in` keyword, and a sequence of values to iterate over.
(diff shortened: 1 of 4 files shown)


meatball133:
The introduction docs for concept exercises also execist under `concepts/concept/about.md` and `introduction,md`. In this case they are located under the concept for-loops. Both of which has to be updated. 

For the Swift track to keep things simple have I made so the introduction docs for the exercise is the same as for the concept (unless an exercise have many concepts). The about file might differ and might have some extra headings the introduction one don't have.

meatball133:
There is this tool which is part of configlet (the cross-track Exercism tool): https://exercism.org/docs/building/tracks/concept-exercises#h-file-docs-introduction-md-tpl. However, it only allows the introduction to be templated. Not the about file.

 Alternatively, you can develop a custom solution (when becoming maintainer). I have done such a thing on the main track I maintain: https://github.com/exercism/crystal/tree/main/document-generator, I haven't worked on it for a while but it does work.

PR #867, opened 2025-07-11
[New Exercise] Add word-search exercise

Sencudra:


config.json
```diff
@@ -1039,6 +1039,17 @@
           "transforming"
         ]
       },
+      {
+        "slug": "word-search",
+        "name": "Word Search",
+        "uuid": "e88da49b-37d7-4f84-ad64-765fb464a721",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
```
meatball133, line 1048: I would increase the difficulty to at least  5

exercises/practice/word-search/Tests/WordSearchTests/WordSearchTests.swift
```diff
@@ -0,0 +1,542 @@
...
+    let words = ["rust"]
+
+    var expected = [String: WordLocation?]()
+    expected["rust"] = nil
+
+    #expect(search(words: words, in: grid) == expected)
+  }
+}
+
+extension WordLocation {
```
meatball133, line 526: I get the idea that putting this in the source file would clutter it quite a bit. 

I think a better approach would be to put this whole struct as a whole in a separate source file. As you can see being done here: https://github.com/exercism/swift/tree/main/exercises/practice/saddle-points/Sources/SaddlePoints.

config.json
Sencudra: Agreed. Done.

PR #868, opened 2025-07-12
Remove Ubuntu from CI workflow, using only macOS for testing

meatball133:
Due to the Ubuntu ci have a tendency to fail (this is due to a dependency which has this as known issue). It is intended to be solved with the next version which seems to be stalled atm. So I decide to disable them for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, macos-15]
+        os: [macos-15]
         swift: ["6.0", "6.1"]
     env:
       RUNALL: "true"


BNAndras:
Can we just use the test runner image inside the c Linux CI? Otherwise, the CI is testing solutions only on macOS but the test runner is Linux. That doesn’t make a lot of sense.

IsaacG:
Is can we use a different Linux distro? Is there some specific issue with the runner we could work around? Could we backport the fix?

IsaacG:
Is there a prior version of that GHA that wasn't broken?

BNAndras:
https://github.com/swift-actions/setup-swift/releases/tag/v2.2.0 predates when the bug was reported in April (https://github.com/swift-actions/setup-swift/issues/694). That wouldn’t support Swift 6.1 which the test runner targets unfortunately.

BNAndras:
Since the issue seems to be getting Swift easily installed, https://hub.docker.com/_/swift/tags?name=6.1 shows several Docker images that would already have it installed.

glennj:
I would suggest that, as this is "just" for CI, not the test runner, we shouldn't be too worried about the efficiency of the infrastructure. (Unless it's easy to add a docker image into a github CI workflow)

meatball133:
> Can we just use the test runner image inside the Linux CI? Otherwise, the CI is testing solutions only on macOS but the test runner is Linux. That doesn’t make a lot of sense.

Prior to the Swift 6 update which was released last month. It was only macOS testing and it worked totally fine. Also worth mentioning is that I use Linux to make the patches (and I think the other quite active contributor also do)  so some Linux testing already occurs. This was mostly intended as a quick temporary fix until the library updates to fix this.

PR #879, opened 2025-09-15
Remove Swift 6.0 from CI matrix, keeping only Swift 6.1

meatball133:
Swift 6 ci is now also broken (Ubuntu was broken and removed). I would suppose that this is caused by the underlying macos image changed which caused it to fail. For now the easiest fix is just removing the ci.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-15]
-        swift: ["6.0", "6.1"]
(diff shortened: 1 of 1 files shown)


Sencudra (approved):

meatball133:
>Looks okay for me. Initially, what was the idea to add both 6.0 and 6.1?

Generally, it is a good idea to support multiple versions, so people aren't forced to use the latest versions. For example Linux distrubtions doesnt always have the latest version of a package.

PR #902, opened 2026-06-28
wrong layout in __inout parameter__ concept introduction: still not fixed

kawazaki42:
image

if you search `exercism/warning` on `github.com/exercism`, it gives just 8 mentions, while searching `exercism/note` and `exercism/caution` gives >100. so i think it just doesn't understand `warning`

yet point 1 isn't very clear that it's about ownership and such, even in context of point 2. it sounds like "you can accept an inout parameter, yet you can't use it" instead of "don't alias the variable and its inout version in the same function". we probably also need an example for it?

### DIFF
--- a/concepts/inout-parameters/about.md
+++ b/concepts/inout-parameters/about.md
@@ -22,7 +22,7 @@ updateVersion(&dbRecord)
 // dbRecord is now (3, "Exercism")
 ```
 
-~~~~exercism/warning
+~~~~exercism/caution
 There are a couple of extra rules one should be aware of regarding in-out parameters.
 
 1.  Inside a function with in-out parameters, you are not allowed to reference the variable that was passed in as the in-out parameter.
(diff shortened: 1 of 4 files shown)


concepts/inout-parameters/about.md
```diff
@@ -22,12 +22,27 @@ updateVersion(&dbRecord)
...
+var directVar = 0
+
+func inoutFunc(_ ioVar: inout Int) {
+  ioVar += 1
+  print(directVar)
+}
+
+inoutFunc(&directVar)
+// raises a compiler error:
+// "main actor-isolated var 'directVar' can not be referenced from a nonisolated context"
```
BNAndras, line 41: I don't speak Swift, but this compiler error doesn't seem to be about `inout`. The example should fail specifically because we're accessing something passed as `inout`. We'll probably want to update this example for every instance it appears in the PR.

BNAndras (changes requested):

IsaacG:
I think this is trying to fix the same thing at #870  ?

